PR #3, opened 2022-01-30
Default system is musl

tomberek:
comparing with the more limited build.system caused a key collision in the attrset


blaggacao (approved):

blaggacao:
Thanks for the catch!

PR #4, opened 2022-01-30
Match upstream

tomberek:
With doCheck enabled, the default package is likely to be built/cached by
cache.nixos.org already.


blaggacao:
Let's educate! :+1:

PR #51, opened 2022-04-15
feat/std tui

blaggacao:
- WIP: std TUI
- WIP: std TUI
- WIP: fix filter

- [x] status with actual run command of any given action -> help to find the corresponding CI command
  eg: `In a script, run: 'std //cell/organelle:target action'`
- [x] spinner during initial load of `__std` metadata
- [x] `?` to render markdown (https://github.com/knipferrc/teacup/blob/main/markdown/markdown.go) info per target in a overlay of https://github.com/charmbracelet/bubbles#viewport (add `__std_readme` meta) using a dialog box:
```go
		dialog := lipgloss.Place(width, 9,
			lipgloss.Center, lipgloss.Center,
			dialogBoxStyle.Render(ui),
			lipgloss.WithWhitespaceChars("猫咪"),
			lipgloss.WithWhitespaceForeground(subtle),
		)
```
- [ ] optional for first iteration: add `std ...` correspondence CLI commands based on `cobra`

Hack this branch:
```
cd cells/std/cli
watchexec -rw . "go mod vendor && go run ."
```


cells/std/cli/targets.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return &TargetModel{targetList}
+}
+
+type item struct {
+	name        string `json:"__std_name"`
+	organelle   string `json:"__std_organelle"`
+	cell        string `json:"__std_cell"`
+	clade       string `json:"__std_clade"`
+	description string `json:"__std_description"`
```
blaggacao, line 71: We could add here:
```
readme string `json:"__std_readme`
```

cells/std/cli/randomitems.go
```diff
@@ -0,0 +1,273 @@
...
+	names          []string
+	nameIndex      int
+	organelles     []string
+	organelleIndex int
+	cells          []string
+	cellIndex      int
+	clades         []string
+	cladeIndex     int
+	descs          []string
+	descIndex      int
```
blaggacao, line 18: and here an appropriate implementarion for generating dummy simple markdowm readmes

blaggacao (commented):
Ideas for setting up readme testdata

cells/std/cli/randomitems.go
```diff
@@ -0,0 +1,273 @@
...
+		"Too ripe",
+		"At last",
+		"What?",
+		"Wow",
+		"Yum",
+		"Maybe",
+		"Sure, why not?",
+	}
+
+	r.shuffle.Do(func() {
```
YorikSar, line 95: r.shuffle has just been unconditionally created in this very function, so it will always execute code in this anonymous function, set "done" flag internally and will never be touched again.

cells/std/cli/randomitems.go
```diff
@@ -0,0 +1,273 @@
...
```
YorikSar, line 229: Same here

cells/std/cli/randomitems.go
```diff
@@ -0,0 +1,273 @@
...
+
+}
+
+func (r *randomActionGenerator) next() action {
+	if r.mtx == nil {
+		r.reset()
+	}
+
+	r.mtx.Lock()
+	defer r.mtx.Unlock()
```
YorikSar, line 249: And also I saw only one place where these generators are called and it's very not threaded, so I don't think locks are needed here.

blaggacao:
Maybe this PR is relevant? https://github.com/charmbracelet/bubbletea/pull/237

PR #61, opened 2022-05-03
Avoid multiple eval of `grow args` in `growOn`

jbgi:
when there is more than one layer of soil.

 Note that this should also allow to merge `grow` and `growOn`
 into a single function if so desired.


blaggacao:
Thanks! -- Why easy when we can have it complicated? 😁

PR #62, opened 2022-05-10
move some commands to optionals

GTrunSec:
making the shell commands clarify.

By the way, could you please add the filter to src of the CLI ? There is no reason to compile again when some code changed.


blaggacao:
Thanks!

blaggacao:
I think you meant optimizing this:
https://github.com/divnix/std/blob/cbb7180c44ca9c5b2b805a863956ad1020e8c55c/cells/std/cli.nix#L12

That's definitely something for a release version to consider

PR #65, opened 2022-05-16
Don't overwrite flake's nixpkgs input

YorikSar:
std used to overwrite nixpkgs input with its own even if it's provided.
Switch to adding nixpkgs input only if it is provided by flake and only with the one provided by flake.


blaggacao:
That makes a lot of sense, thank you

michalrus:
Yes, a lot of sense, thank you!

PR #66, opened 2022-05-20
Add support for setting nixpkgs overlays.

jbgi:


blaggacao:
In fact, `std` tries to discourage nixpjgs overlays, because in my analysis, the codimg oattern that go with global overlays are one of the many reasons why nix is hard on newcomers.

So let's keep discouraging them for the sake of local reasoning.

blaggacao:
Overlays can still be easily used locally with `nixpkgs.extend`.

By not making them available globally, the following problem is solved:
- unchecked hidden context that make it very hard to reason about the repo

By encouraging the local use, we gain:
- local reasoning
- increased eval time / ram (if used extensively)

roberth:
An alternative, general way to have overlays is to _actually_ support them in the `nixpkgs` flake, instead of reinventing the wheel with `import nixpkgs { ..... }`. https://github.com/NixOS/nixpkgs/pull/160061 solves this for `overlays`, `config` and cross compilation (setting the build `system` while still taking the host `system` from the flake `${system}` attribute name).

The design objections against overlays of course still stand, and I will join y'all to recommend to avoid them when possible.

PR #70, opened 2022-05-26
fix vendorSha

GTrunSec:
Do we need a simple github CI check?


blaggacao:
> Do we need a simple github CI check?

That would be great! :-)

PR #71, opened 2022-05-31
fix(docs): typo in incl.nix

phaer:
Hi, thanks for this interesting project! I just stumbled upon a few probable typos while reading and thought I'll send a quick PR.


blaggacao:
Thanks, I'm not officially a legasthenic, but I probably should apply for. :smile:

PR #78, opened 2022-06-14
deSystemize nixpkgs instead of instantiating it

YorikSar:
Essentially replace nixpkgs input with deSystemized legacyPackages in it.
This allow to use packages for fixed system, on any host system.
For example, in Docker images we should always use x86_64-linux, even if we're building images on macOS.


blaggacao:
Thank you! Merged with a small fixup in https://github.com/divnix/std/commit/f9c70514397b99e0f97567e9b48981abcab319f8

PR #79, opened 2022-06-16
fix mdbook-krioki command

GTrunSec:
It seems we can not render the table/list when mdbook-kriki enabled.


blaggacao:
Thank you!

PR #82, opened 2022-06-21
imp: add container clade (nix2container)

blaggacao:
- This is a pretty common devops scenario (build & push an OCI-image)
- We are nice. Hence, we cover it ootb


blaggacao:
Tested manually and works as expected

PR #85, opened 2022-06-23
Extend the installables clade

GTrunSec:


blaggacao:
Reportedly works as intended.

PR #86, opened 2022-06-27
imp: add standardized writeShellEntrypoint

blaggacao:
- implemets an interface to produce OCI-images with the following
  contracts:

```
/bin/entrypoint
/bin/debug
/bin/live
/bin/ready

```

### DIFF
--- a/cells/std/lib/writeShellEntrypoint.md
+++ b/cells/std/lib/writeShellEntrypoint.md
@@ -43,6 +43,7 @@ The following contracts can be consumed:
 
 ```
 /bin/entrypoint # always present
+/bin/runtime    # always present, drops into the runtime environment
 /bin/live       # if livenessProbe was set
 /bin/ready      # if readinessProbe was set
 ```
@@ -63,6 +64,7 @@ The following contracts can be consumed:
 
 ```
 /bin/entrypoint # always present
+/bin/runtime    # always present, drops into the runtime environment
 /bin/debug      # always present, drops into the debugging environment
 /bin/live       # if livenessProbe was set
 /bin/ready      # if readinessProbe was set
--- a/cells/std/lib/writeShellEntrypoint.md
+++ b/cells/std/lib/writeShellEntrypoint.md
@@ -1,0 +1,105 @@
+# `writeShellEntrypoint`
+
+... is a function to write Standard OCI-image entrypoints.
+
+The function signature is as follows:
+
+```nix
+{
+  # the installable that is wrapped by this entrypoint (re-exported)
+  package,
+  # the literal bash string of the entrypoint that will be wrapped
+  entrypoint,
+  # initialize environment variables with these defaults
+  env ? {},
+  # runtime installables that the entrypoint or liveness/readiness probe uses (re-exported)
+  runtimeInputs ? [],
+  # domain specific debugging utilities (re-exported)
+  debugInputs ? [],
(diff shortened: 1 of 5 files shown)


cells/std/lib/writeShellEntrypoint.nix
```diff
@@ -0,0 +1,212 @@
...
+    inner =
+      nixpkgs.writeTextFile {
+        name = "entrypoint";
+        executable = true;
+        destination = "/bin/entrypoint";
+        inherit checkPhase;
+        text = ''
+          ${prelude}
+
+          [ -n "''${DEBUG_SLEEP:-}" ] && sleep "$DEBUG_SLEEP"
```
manveru, line 179: maybe `sleep "${DEBUG_SLEEP:-0}"` instead? Otherwise using `[[ ... ]]` would avoid the need to call `test` and use the bash builtin instead.

cells/std/lib/writeShellEntrypoint.nix
```diff
@@ -0,0 +1,212 @@
...
+
+  mkDebugOCI = entrypoint: name: let
+    debug-banner = nixpkgs.runCommandNoCC "debug-banner" {} ''
+      ${nixpkgs.figlet}/bin/figlet -f banner "STD Debug" > $out
+    '';
+    debug-tools = with nixpkgs.pkgsStatic; [
+      bashInteractive.out
+      busybox
+    ];
+    debug-bin = nixpkgs.writeShellApplication {
```
manveru, line 55: This here means we now have three versions of bash in the image... one is just `bash`, then `pkgsStatic.bash`, and `pkgsStatic.bashInteractive`.

blaggacao (commented):
You know me.

cells/std/lib/writeShellEntrypoint.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Contracts
+
+The following contracts can be consumed:
+
+```
+/bin/entrypoint # always present
+/bin/live       # if livenessProbe was set
+/bin/ready      # if readinessProbe was set
+```
```
blaggacao, line 48: `/bin/runtime`

cells/std/lib/writeShellEntrypoint.nix
```diff
@@ -0,0 +1,216 @@
...
+  mkDebugOCI = entrypoint: name: let
+    debug-banner = nixpkgs.runCommandNoCC "debug-banner" {} ''
+      ${nixpkgs.figlet}/bin/figlet -f banner "STD Debug" > $out
+    '';
+    debug-tools = with nixpkgs.pkgsStatic; [busybox];
+    debug-bin = nixpkgs.writeShellApplication {
+      name = "debug";
+      runtimeInputs =
+        (entrypoint.runtimeInputs or [])
+        ++ entrypoint.debugInputs or []
```
disassembler, line 56: doesn't including debugInputs in runtimeInputs defeat the purpose of creating an image that doesn't have debugInputs included?

blaggacao: This is the `/bin/debug` package, so that is absolutely intended.

blaggacao: This package is only present in `mkDebugOCI`, though.

blaggacao:
@disassembler Is there still some outstanding polishing coming from our first implementation exercise? I'd like to get this merged as a dependency for an upcoming PR...

PR #87, opened 2022-06-27
imp: spellcheck

johnalotoski:


nrdxp (approved):
Nice job! Did you use a tool for spellcheck? If so we might wanna add it to the pre-commit hook. Even so we can do that in a follow up PR.

johnalotoski:
Thanks!  I started noticing them while just reading the docs and decided to fix as I was reading, so enabled vim spellcheck to catch more.  Then piped all md files through aspell at the end to see if I missed any in vim and caught another couple.  Pre commit hook is a nice idea, but there are a lot of false positives with all the technical terms intermingled, so might be a bit tough.

blaggacao:
I'm glad to have your pair of eyes on this (and above all what's between/behind them). Thank you!

PR #91, opened 2022-07-11
direnv: improve load speed and address shellcheck

nrdxp:
There is still an issue where the time for a `nix build` is non-deterministic and takes far longer about 1/10 times, even if the evaluation is already fully cached, which will likely require a good solution for NixOS/nix#2913.

Since the issue is out of scope for `std`, I'm opening this PR anyway.


blaggacao (approved):
Thank you! LGTM

PR #97, opened 2022-07-26
imp: add initial version of a nixago clade

blaggacao:
Still draft due to dependency on:
- https://github.com/nix-community/nixago/pull/41
- https://github.com/nix-community/nixago/pull/42


cells/std/lib/default.nix
```diff
@@ -4,7 +4,61 @@
...
+      in {
+        options.nixago = mkOption {
+          type = types.listOf types.attrs;
+          default = [];
+          apply = x: x.passthru;
+          description = "List of `std` Nixago pebbles to load";
+        };
+
+        config.devshell = mkIf (cfg.nixago != []) {
+          startup.nixago-setup-hook = stringBefore ["motd"] inputs.nixago.lib.makeAll cfg.nixago;
```
blaggacao, line 24: @jmgilman One issue here, that becomes immediatly evident is that `nixago`'s common hook doesn't cleanup after itself, now does it preserve eventual competing funtion definions of e.g. `debug`.

I'm not entirely sure what's the best way to preserve the original functions if they exist and then only temporarily load the nixago ones, `unset -f ` them at the end and restore the originals...

blaggacao: for context, we talk about these: https://github.com/nix-community/nixago/blob/master/lib/hooks/common.nix

zeme-iohk (approved):

GTrunSec (approved):

blaggacao:
Waiting on these two, still:

- https://github.com/nix-community/nixago/pull/44
- https://github.com/nix-community/nixago/pull/45

Otherwise: ready!

cells/std/lib/default.nix
blaggacao: fixed in https://github.com/nix-community/nixago/pull/46

blaggacao:
Added an ADR for Nixago & Docs. Off we go!

PR #100, opened 2022-07-30
add missing clade

GTrunSec:


GTrunSec:
@blaggacao  How about only putting the `clades` in flake.nix? That would be more clear for use like `std.clades.data <org>`.

blaggacao:
Thanks for this fixup!

Yep, we could do that, also in the perspective that there might be more and more clades and also for the sake of transparent naming on the calling site.

PR #113, opened 2022-08-29
Add setup-hook to update the direnv_lib.sh automatically.

GTrunSec:


blaggacao:
This is a nice trick!

Maybe you have not yet seen this recent commit, which completely and fundamentally does away with this problem thanks to the power of nix itself.
 https://github.com/divnix/std/commit/2841945d146dbb218deef0c96a01a7336ac74b8c

Apologies for the inconveniences this might have created.in the past.

GTrunSec:
Oh, pretty elegant solution. thanks let me close it.

PR #119, opened 2022-09-03
feat: adds just nixago pebble

### LINKED ISSUE #118: Build a bridge into `just` (via nixago)
Build bridges.

Just is familiar and can serve as an on-ramp for the use of nixified open source repos that also want to appeal to non-nix.

It's a little less prevalent than `make`, so we can teach the benefits of `nix shell nixpkgs#just` alongside until the user is ready to cut the learning bridge indirection and go for the real stuff.

We need a Nixago pebble that automatically converts some sort of "repo task best practice according to Standard" into a `Justfile`.

It is ok, if that `Justfile` can't be run reproducibly, ootb: that's just another teaching opportunity. Within the Devshell, it should be reproducible, though.

jmgilman:
Fixes #118


.gitignore
```diff
@@ -8,6 +8,7 @@
 result
 
 # nixago: ignore-linked-files
+.justfile
```
blaggacao, line 11: I would typically not link it. The "on-ramp" can only work if that familiar file is there also when the user looks at Github.

That's actually the only (non-technical) reason to copy it, instead.

cells/std/nixago.nix
```diff
@@ -57,6 +57,23 @@ in
...
+        files = [./nixago/just.cue];
+        flags = {
+          expression = "rendered";
+          out = "text";
+        };
+        postHook = ''
+          cat $out
+          ${inputs.nixpkgs.just}/bin/just --unstable --fmt -f $out
+        '';
+      };
```
blaggacao, line 75:
```suggestion
      };
      apply = d: {data = d;};
```

cells/automation/nixago.nix
```diff
@@ -34,6 +34,19 @@ in {
...
+      data = {
+        tasks = {
+          fmt = [
+            ''
+              treefmt $(git diff --name-only --cached)
+            ''
+          ];
+        };
+      };
+    };
```
blaggacao, line 48:
```suggestion
    configData = {
      tasks = {
        fmt = [
          ''
            treefmt $(git diff --name-only --cached)
          ''
        ];
      };
    };
```

blaggacao (changes requested):
Here is how we can make the `configData` iface a little more intuitive.

cells/std/nixago/just.cue
```diff
@@ -0,0 +1,29 @@
...
+    head: data.head
+    tasks: {
+        for task, steps in data.tasks {
+            "\(task)": strings.Join(steps, "\n    ")
+        }
+    }
+}
+
+tmpl:
+"""
```
blaggacao, line 21:
```suggestion
"""
# Autogenerated! DO NOT EDIT!
# This justfile is autogenerated via https://divnix.github.io/std/reference/std/nixago/just.html
# So that you can run the fundamental repository tasks with your favorit task runner.
# No need to install Nix, for now.
```

blaggacao: https://divnix.github.io/std/reference/std/nixago/just.html would have to be made a thing, as well. :smile:

blaggacao (commented):
Let's write something that a reader can interpreted like an "on-ramp" to Nix.

This is just an idea, but I'm sure this can be transported way better...

GTrunSec:
looks good. I have one suggestion that adds a description, which should be placed above the task in the template. 

- will present a description info like devshell.command.description
0:2:zsh -  NixOS   2022-09-03 16-35-52


- the nix part will be changed like:

```nix
task = {
  fmt = {
  description = "format your code with treefmt";
  content = ''
  <command>
   '';
  };
}
```
- the justfile part:
```
# format your code with treefmt                                                                                        
   fmt :
      <command>
```

jmgilman:
Funny enough, I have this change already implemented just not committed yet :) I'm attempting to perform this change in tandem with integrating with the devshell `menu`.

GTrunSec:
Did you consider the Args flag? it's pretty simple. just add the args after the task name.

["arg1" "arg2"]

'''
fmt Arg1 Arg2:

jmgilman:
This is ready to ship, The only thing we might consider throwing in is devshell integration, but that seems like a hard nut to crack.

GTrunSec:
I don't recommend integrating just related commands in the devshell menu. It will make the menu bloated. In the real user cases, just and command-completion have already had an outstanding experience. Once we find a way to hook the `nix devshell` into a [nutshell](https://www.nushell.sh/), I believe there will be a better experience with menu extensions.

GTrunSec (approved):

jmgilman:
> Once we find a way to hook the `nix devshell` into a [nutshell](https://www.nushell.sh/), I believe there will be a better solution with menu extensions.

What are your ideas on the nushell integration?

blaggacao:
> Once we find a way to hook the nix devshell into a nutshell,

That would be a bold move all together. 😁

https://github.com/nushell/nu_scripts/tree/main/direnv

But it looks like nushell is a strictly better shell, indeed.

PR #127, opened 2022-09-09
harvest: derive harvest from more general function 

nrdxp:
After starting #126, I realized all that was needed was to modify harvest to allow for skipping marked derivations based on a predicate. Winnow is the same function as harvest, except it accepts a filter as it's first argo.


blaggacao (approved):
Super nice and clean! Thank you!

PR #129, opened 2022-09-10
refactor(nixago): moves optional nixago configs to presets cell

### LINKED ISSUE #128: We need a layer between presets and their boilerplate.
Users of std _want_ batteries included. But also removable.

The batteries that currently come with `std.std.nixago` are not removable (e.g. use `alejandra` as the Nix formatter).

Introduce a `std.presets` cell that is dedicated to batteries whenever that distinction is appropriate.

jmgilman:
Fixes #128.


jmgilman:
The *.md files for the Nixago configs seem redundant with the [documentation already provided for them](https://divnix.github.io/std/reference/std/nixago/index.html). Can we consider removing these? It cleans up the directory and makes it easier to find the *.nix files.

jmgilman:
One other thing. With the way [this is setup](https://github.com/divnix/std/blob/3335a48d421604093e5e48527e870973d9735511/cells/std/lib/default.nix#L47-L52), we have to do [things like this](https://github.com/divnix/std/blob/main/cells/std/nixago.nix#L61). I wonder if we just consider making `config` optional here is well? It would reduce a bit of boilerplate.

cells/presets/nixago.nix
```diff
@@ -0,0 +1,61 @@
...
+    configData = import ./nixago/adrgen.nix;
+  };
+  editorconfig = std.nixago.editorconfig {
+    configData = import ./nixago/editorconfig.nix;
+    hook.mode = "copy"; # already useful before entering the devshell
+  };
+  conform = std.nixago.conform {
+    configData = import ./nixago/conform.nix;
+    format = "yaml";
+    output = ".conform.yaml";
```
blaggacao, line 19: Is this copying over on purpose? If so, I'm not sure I understand why here and not with the others. I'd expect _not_ to redefine it here.

blaggacao (changes requested):
`mdbook.hook.extra` in `cells/std/nixago.nix` appears to still consume a contract from the now moved `cells/std/nixago/mdbook.nix`

blaggacao:
> The *.md files for the Nixago configs seem redundant with the documentation already provided for them. 

These are symlinks. In their current place, these docs are also rendered inside the TUI, when you type`?`.

We should likely just leave them here, since they do little more than link to the tools with a one-liner decsription.

blaggacao:
> I wonder if we just consider making config optional here is well? It would reduce a bit of boilerplate.

Since this is sort of boilerplate within boilerplate, I think it actually conveys a bit of information, namely that it's not _actually_ optional. Well, technically, it is, but one would end up with an empty file.

So once one is actually looking at the boilerplate, one can still draw value of it.

So I'd leave it as it is, here.

blaggacao:
Thanks a lot for this cleanup. :rocket: :+1:  I just rebased and merged into master here: https://github.com/divnix/std/commit/bff0e737b872af59ba70fdf6e06b7dc212156d17

closing.

PR #132, opened 2022-09-12
imp: implement `nomadJobsManifest` block type

nrdxp:
~~Will render and check for correctness each Nomad job in the given env (Nomad namespace)~~ Now only targets single jobs.

This will not allow jobs to be built from a dirty tree, so that deployments can always been tracked back to a commit. 

~~As a potential improvement, we could also potentially inject the commit into the job meta, to make the build commit easily discoverable from the running job description. Just haven't worked out the logic for that yet.~~ implemented: 6c790dae788e24670edd9deed25a23317e155190

TODO:
- [x] deploy action
- [x] explore action


blaggacao:
superseded by https://github.com/divnix/std/pull/136 (see PR description)

PR #144, opened 2022-09-16
feat: allows additional options to be passed to mkOCI/mkDebugOCI

jmgilman:
I'm not sure how to make this backward compatible. The problem is that the current structure blocks access to arguments being passed to `n2c.buildImage`. The most flexible option seems to convert the `mkOCI` functions to accept an attribute set that takes an optional `options` argument for merging against whatever the internal `mkOCI` function does. 

With this change, I can do this now:

```nix
{
  default = entrypoint.mkOCI {
    name = "docker.io/cocogitto-pr";
    options = {
      tag = "latest";
      config.Labels = {
        "org.opencontainers.image.title" = "cocogitto-pr";
        "org.opencontainers.image.version" = "0.1.0";
        "org.opencontainers.image.url" = "https://github.com/jmgilman/cocogitto-pr";
        "org.opencontainers.image.source" = "https://github.com/jmgilman/cocogitto-pr";
        "org.opencontainers.image.description" = ''
          Github Action for generating preview PRs with cocogitto
        '';
      };
    };
  };
}
```

### DIFF
--- a/cells/std/lib/writeShellEntrypoint.nix
+++ b/cells/std/lib/writeShellEntrypoint.nix
@@ -204,8 +204,8 @@
   in
     inner
     // {
-      mkOCI = name: n2c.buildImage (mkOCI inner name);
-      mkDebugOCI = name: n2c.buildImage (mkDebugOCI inner name);
+      mkOCI = { name, options ? {}}: l.recursiveUpdate (n2c.buildImage (mkOCI inner name)) options;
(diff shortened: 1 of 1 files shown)


jmgilman:
Another issue is that `mkOCI` assumes we want to put the entrypoint into `config.Cmd` and offers no way to put it into `config.Entrypoint`, even with the current PR (because I cannot access the entrypoint derivation after its created).

jmgilman:
Closed in favor of #146

PR #146, opened 2022-09-22
feat: adds mkOperable and updated version of mkOCI w/ deps

jmgilman:
This PR introduces a few things.

## Entrypoint -> Operable

See: [The four layers](https://divnix.github.io/std/patterns/four-packaging-layers.html#the-layers)

Previously, a package would be turned into an _entrypoint_. This was necessary to transform a package, which may or may not be "operable", into a self-contained derivation that included everything required to operate the package (runtime inputs, environment variables, etc.). The problem is that _entrypoint_ is universally associated with containers, and thus the _entrypoint layer_ risks unnecessarily bleeding into a downstream layer.

This PR begins to change _entrypoint_ to _operable_ to distinguish the transformation better.

## mkOperable

Concretely implements the above change. Returns an "operable" derivation for a given package. It also includes optional liveness and readiness probes which can pass thru to later layers. 

## mkOCI and deps

A more robust version of the current `mkOCI` implementation. Transforms an operable into an OCI image. It brings the following additional benefits:

- Offers more tuning options to better control the final image (i.e. tag, labels, etc.).
- Adds support for _setup tasks_ (via `mkSetup`) that allows arbitrarily modifying the root filesystem of the container.
- Allows controlling the running user, including adding a new user to the image with the `mkUser` setup task.
- Additional layer optimizations to reduce overall image size.
- Adds a catch-all `options` argument for augmenting the final arguments sent to `n2c.buildImage`

Note that this PR is dependent on [this PR](https://github.com/nlewo/nix2container/pull/47) which adds support for modifying file ownership. It should not be merged before that one.

### DIFF
--- a/cells/std/lib/default.nix
+++ b/cells/std/lib/default.nix
@@ -80,14 +80,16 @@ in {
     __functor configuration' {};
 
   fromMakesWith = inputs': let
-    inputsChecked = assert l.assertMsg (builtins.hasAttr "makes" inputs') (
+    inputsChecked = assert l.assertMsg (builtins.hasAttr "makes" inputs')
+    (
       l.traceSeqN 1 inputs' ''
 
         In order to be able to use 'std.std.lib.fromMakesWith', an input
         named 'makes' must be defined in the flake. See inputs above.
       ''
     );
-    assert l.assertMsg (builtins.hasAttr "nixpkgs" inputs') (
+    assert l.assertMsg (builtins.hasAttr "nixpkgs" inputs')
+    (
       l.traceSeqN 1 inputs' ''
 
         In order to be able to use 'std.std.lib.fromMakesWith', an input
@@ -95,7 +97,8 @@ in {
       ''
     ); inputs';
     makes = l.fix (
-      l.extends (
+      l.extends
+      (
         _: _: {
           inherit (inputsChecked.nixpkgs) system;
           inputs = inputsChecked;
@@ -114,7 +117,8 @@ in {
   fromMicrovmWith = import ./fromMicrovmWith.nix {inherit nixpkgs;};
 
   writeShellEntrypoint = inputs': let
-    inputsChecked = assert nixpkgs.lib.assertMsg (builtins.hasAttr "n2c" inputs') (
+    inputsChecked = assert nixpkgs.lib.assertMsg (builtins.hasAttr "n2c" inputs')
+    (
       nixpkgs.lib.traceSeqN 1 inputs' ''
 
         In order to be able to use 'std.std.lib.writeShellEntrypoint', an input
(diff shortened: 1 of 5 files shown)


nrdxp:
wouldn't runtime dependencies be considered a packaging layer concern?

jmgilman:
> wouldn't runtime dependencies be considered a packaging layer concern?

I suppose it depends on where the boundary is drawn. A package could encompass everything that the _operable_ does and we could then just eliminate the layer entirely. The only distinction I think that could be made is that a package would only contain the dependencies required to compile it, and any other runtime dependencies would be included in the operable.

But yes, to your point, Nix does not really make that distinction by default.

nrdxp:
> Nix does not really make that distinction by default.

Well not explicitly, but implicitly any string contexts still present in the final out path are considered "runtime" dependencies of a derivation and are kept as part of the runtime closure, which is how dynamic library depends are managed. In terms of executable runtime deps, it is typically good practice to wrap the binary to include them in the path.

I have always sort of felt that the "entrypoint" layer is a tad superflous, but I suppose if the goal is to keep the final runtime scheduler agnostic, it might not be such a bad idea. I also like your probes addition. I'm just trying to distinguish in my mind where the packaging layer ends and this one begins, it seems to be the fuzziest boundary to me.

jmgilman:
> In terms of executable runtime deps, it is typically good practice to wrap the binary to include them in the path.

Yeah, this is where I think the line is currently drawn, hence "runtime inputs" being equivalent to dependent binaries. For example, if I package a bash script that relies on git, it'd be put into the runtime inputs. However, it does make more sense for it to be an explicit dependency of the package itself since the script wouldn't really work without it (and a consumer might be a bit surprised). 

For what it's worth, I don't think the operable layer is always strictly necessary. It's really just a glorified bash script that does whatever is necessary to make the package operate (this usually only makes sense in long-running services). If we move the runtime inputs back into the packaging layer, then we're left with the probes, environment variables, and bash script. I guess the question then becomes: is it worth it to dedicate a layer to these things still?

blaggacao:
>  I guess the question then becomes: is it worth it to dedicate a layer to these things still?

Yes. I want to encourage adding the dynamic perspective in the analysis. As time goes by and a project progresses, everything stashed in the operable layer, probably should migrate into their native scopes.

The runtime dependencies is an example that really should migrate into the packaging layer, another example are missing app capabilities that close the gap between a "runnable" and a proper "operable" (-> "the glorified bash script").

Hence, we can state this as a dynamic boundary around a problem spaces that increasingly approaches nil as the project progresses. This is actually a decent mapping to real project dynamics and roles. And hence the pattern piece on the four layers mentions:

> The role of this script — customarily written in bash — serves as a concise and reified communication channel between Developers and Operators. As such, Operators will find all the primary configuration options re-encoded at a glance and in a well-known location. In the other direction, Developers will find all the magic ad-hoc wrapping that Operators had to engage in, in order to run the application on the target scheduler.

jmgilman:
We should be at feature parity with the original implementation now. Let me know if I've missed anything.

jmgilman:
Ok, now it _really_ has parity 😉

jmgilman:
I'm wondering if it's worth moving the runtime/debug shells back into the operable layer. These are really just "runnables" that bring the environment/runtime inputs into scope. I could imagine this being useful for alternate options in the third layer (i.e., firecracker microvm).

jmgilman:
Ok, I moved the runtime/debug shells to the operable layer. I also generalized `mkOCI` and added a separate function for explicitly covering an operable to an OCI. I'm happy to walk this one back, but I think it's necessary to allow more flexibility in the future (i.e. making an OCI from a devshell).

PR #147, opened 2022-09-24
feat: allow implicit block API

nrdxp:
Cell Blocks always have the same API (i.e. `{inputs, cell}:`).

With this addition any cell block can simply assume that `inputs` & `cell` are in scope without having to explicitly declare them, removing the little bit of boilerplate that we did have.

For backwards compatibility, the old API and its validation is maintained if the file contains a lambda.

Before:
```nix
{inputs, cell}: let
  inherit (inputs.cells.myCell) packages; # redundant ref to inputs
in {
  inherit (packages) myPackage;
}
```

After:
```nix
let
  inherit (inputs.cells.myCell) packages; 
in
{
  inherit (packages) myPackage;
}
```

### DIFF
--- a/src/grow.nix
+++ b/src/grow.nix
@@ -153,8 +153,6 @@
         cPath = paths.cellPath cellsFrom cellName;
         loadCellBlock = cellBlock: let
           oPath = paths.cellBlockPath cPath cellBlock;
-          importedFile = validate.FileSignature oPath.file (import oPath.file);
-          importedDir = validate.FileSignature oPath.dir (import oPath.dir);
           # minimum data for initializing TUI / CLI completion
           extractInitMeta = name: target: let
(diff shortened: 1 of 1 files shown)


blaggacao:
One of the _major_ goals of `std` is to make things more explicit and hold up the ideal of local reasoning.

It is generally a framework that attempts to make deciphering of nix code (even if only) marginally more straight forward for newcomers.

So it is in-scope to carefully trade off author QoL against that goal.

In this case, my judgement is that this proposed change can't be brought in alignment with these goals.

nrdxp:
I thought you might say something like that, but it's really no different than in most other programming languages where project local modules are in scope and importable. Also, the explicit route is still available, and should probably remain so. In any case, the boilerplate is minimal enough that it isn't a huge deal either way.

PR #153, opened 2022-10-04
patterns

blaggacao:
- feat: adds mkOperable and updated version of mkOCI w/ deps
- refactor: moves debug to dedicated script
- fix: use runCommandNoCC
- feat: adds writeScript to standardize shell scripts
- refactor: moves debug inputs to operable layer
- feat: adds debug sleep to operable layer
- fix: check for liveness/readiness attributes
- fix: remove bash from debug inputs
- feat: allow setting runtimeShell in operable layer
- fix: makes runtime shell propogate
- refactor: only sleep when debug is enabled
- fix: removes redundant operable layer
- fix: removes redundant debug layer
- feat: set -x when debug is enabled
- feat: adds /bin/runtime
- fix: fixes scoping issues
- refactor: moves debug/runtime envs to operable layer
- refactor: separates mkOCI to mkOpOCI and mkOCI
- refactor: moves setup perms to mkOCI
- refactor: allow multiple permissions for mkSetup


jmgilman:
LGTM

PR #157, opened 2022-10-07
feat: adds mkDevOCI for generating devcontainers

jmgilman:
This PR introduces support for generating container images that are compatible with [devcontainers](https://code.visualstudio.com/docs/remote/containers). It ingests a devshell and builds a compatible OCI image which can then be invoked as a devcontainer.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,5 +1,5 @@
 {
-    "image": "std-dev",
+    "image": "std-vscode",
     "settings": {
         "terminal.integrated.shell.linux": "/bin/bash"
     }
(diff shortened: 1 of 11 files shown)


jmgilman:
The last commit was done entirely within the devcontainer for `std`! 🚀

jmgilman:
The function has been generalized so as to not rely on vscode. By default, it creates a "generic" container which can be used like so:

```
docker run -it -v "$(pwd):/work" --entrypoint /bin/bash std-dev
```

If the `vscode` argument is set to `true`, then it will augment the container to be compatible with vscode:

```
devcontainer open .
```

blaggacao (approved):
Minors...

PR #162, opened 2022-10-11
fix: lazier loading of cells blocks.

jbgi:
Allows attribute names from a given block to depends on another block,
 without triggering an infinite recursion error.


blaggacao (approved):
Oh, what a sharp eye!

PR #166, opened 2022-10-19
keep nixpkgs.soureInfo when deSystemized

GTrunSec:


nrdxp (approved):

blaggacao:
This is cool and probably more idomatic than `import nixpkgs.path` (which is not well-known and  `nixpkgs`-specific)

PR #169, opened 2022-10-21
feat: adds mkOperableScript for generating runtime scripts

jmgilman:


blaggacao:
LGTM, but: we also should leave some room for anti-DRY plain prose readability in our code bases. As a caveat.

PR #170, opened 2022-10-22
fix(presets/lefthook): fail on treefmt changes

jbgi:
Otherwise commit can go through even though staged file are modified by `treefmt`.

Also removed unnecessary glob '*'.


blaggacao:
Perfect! Thank you!

PR #171, opened 2022-10-22
impl: `nixpkgsConfig`

nrdxp:
The docs mentioned that the special `nixpkgs` input could be configured with `nixpkgsConfig`, but no such option existed.


blaggacao (approved):
Thanks!

PR #174, opened 2022-10-28
feat(presets/lefthook): allow WIP and fixup! commit

jbgi:
workaround https://github.com/siderolabs/conform/issues/225

 Also people very commonly use WIP commits before tidying up, so
 allowing them is probably necessary if we want people not disabling
 conform pre-commit hook.

 Note that this does not affect CI that should ideally still fail on
 WIP commits (TODO: run conform on CI).


blaggacao (approved):
Really cool solution, thank you!

PR #178, opened 2022-11-03
fix: restore lazy cell loading

jbgi:
ci register feature induced strict evaluation of cell
 blocks, due to lists being strict in nix.

### DIFF
--- a/src/grow.nix
+++ b/src/grow.nix
@@ -131,7 +131,7 @@
           // (
             if tail == [null]
             then {inherit (acc) ci;}
-            else {ci = acc.ci ++ (l.flatten tail);}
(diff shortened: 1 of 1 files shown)


blaggacao:
I honestly felt I had to ask you, because I went suspicious as all of a sudden en error was thrown too early for my taste.

So that you did the tweaking already is just awesome! Thank you!

blaggacao:
Waiting for the linux check to finish ...

blaggacao:
I was almost a bit more concerned about https://github.com/divnix/std/blob/main/src/grow.nix#L277

`mapAttrsToList`. But maybe that's actually ok?!?

PR #184, opened 2022-11-14
feat: add std logo generator

blaggacao:


blaggacao:
Looks good enough for making logos with fonts: https://github.com/paulzi/svg-text-to-path

PR #187, opened 2022-11-17
fix: import nixpkgs from inputs instead

GTrunSec:
that's why we got two nixpkgs
1:3:bash -  NixOS   2022-11-16 21-33-10


blaggacao (approved):

blaggacao:
Thank you!

Issue #188, opened 2022-11-18
editorconfig `root = true` setting breaks `std.nixago.editorconfig`


montchr:
I'm in the process of setting up a repo with Standard and converted my usual `.editorconfig` file into the equivalent `editorconfig.nix` file. EditorConfig supports a `root = true` setting at the top-level of `.editorconfig`, but when I added this, evaluation fails.

`nix flake check` output:

```console
...
error: value is a Boolean while a set was expected

       at /nix/store/jy6lx7safmcgwf1sli6a18rgimfhnm2s-source/lib/attrsets.nix:356:39:

          355|   mapAttrsToList = f: attrs:
          356|     map (name: f name attrs.${name}) (attrNames attrs);
             |                                       ^
          357|
```

<details>
<summary>Full trace output</summary>

```console
error: value is a Boolean while a set was expected

       at /nix/store/jy6lx7safmcgwf1sli6a18rgimfhnm2s-source/lib/attrsets.nix:356:39:

          355|   mapAttrsToList = f: attrs:
          356|     map (name: f name attrs.${name}) (attrNames attrs);
             |                                       ^
          357|

       … while evaluating 'mapAttrsToList'

       at /nix/store/jy6lx7safmcgwf1sli6a18rgimfhnm2s-source/lib/attrsets.nix:355:23:

          354|   */
          355|   mapAttrsToList = f: attrs:
             |                       ^
          356|     map (name: f name attrs.${name}) (attrNames attrs);

       … from call site

       at /nix/store/jy6lx7safmcgwf1sli6a18rgimfhnm2s-source/lib/generators.nix:90:52:

           89|         else k: v: [ (mkLine k v) ];
           90|   in attrs: libStr.concatStrings (lib.concatLists (libAttr.mapAttrsToList mkLines attrs));
             |                                                    ^
           91|

       … while evaluating anonymous lambda

       at /nix/store/jy6lx7safmcgwf1sli6a18rgimfhnm2s-source/lib/generators.nix:90:6:

           89|         else k: v: [ (mkLine k v) ];
           90|   in attrs: libStr.concatStrings (lib.concatLists (libAttr.mapAttrsToList mkLines attrs));
             |      ^
           91|

       … from call site

       at /nix/store/jy6lx7safmcgwf1sli6a18rgimfhnm2s-source/lib/generators.nix:128:14:

          127|           [${mkSectionName sectName}]
[...]

PR #194, opened 2022-11-27
fix(std): nixago editorconfig special ini format

blaggacao:
fix: #188

### DIFF
--- a/cells/std/nixago/editorconfig.nix
+++ b/cells/std/nixago/editorconfig.nix
@@ -2,10 +2,19 @@
   inputs,
   cell,
 }: let
+  l = nixpkgs.lib // builtins;
   inherit (inputs) nixpkgs;
 in {
   configData = {};
   output = ".editorconfig";
-  format = "ini";
+  engine = request: let
+    inherit (request) configData output;
+    name = l.baseNameOf output;
+    value = {
+      globalSection = {root = configData.root or true;};
+      sections = l.removeAttrs configData ["root"];
+    };
+  in
+    nixpkgs.writeText name (l.generators.toINIWithGlobalSection {} value);
   packages = [nixpkgs.editorconfig-checker];
 }
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,5 @@
+root=true
+
 [*]
 charset=utf-8
 end_of_line=lf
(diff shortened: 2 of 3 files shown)


cells/std/nixago/editorconfig.nix
```diff
@@ -2,10 +2,19 @@
...
   inherit (inputs) nixpkgs;
 in {
   configData = {};
   output = ".editorconfig";
-  format = "ini";
+  engine = request: let
+    inherit (request) configData output;
+    name = l.baseNameOf output;
+    value = {
+      globalSection = {root = configData.root or false;};
```
blaggacao, line 14: If not set, defaults explicitly to `false` which is cool, I guess.

montchr: I think `true` would be a better default here.

In my experience across many repos, a proverbial 98% of them only have a single `.editorconfig` file at the project root, where `root = true` would apply. Even in monorepos including a few different languages and sub-projects with their own distinct scope, each of these scopes will likely adhere to the same coding standards declared in the root-level `.editorconfig`. So, in those proverbial 2% of projects where we've had multiple of these files, the additional files turned out to be an unnecessary source of confusion in the end.

cells/std/nixago/editorconfig.nix
```diff
@@ -2,10 +2,19 @@
...
+      sections = l.removeAttrs configData ["root"];
```
montchr, line 15: Interesting -- this reminds me of an oddity I've noticed in a `treefmt.toml` generated with Standard:

```toml
[formatters] # <- unnecessary, but harmless (i think)
[formatters.nix]
command = "alejandra"
...
```

That said, the additional section could have been something I introduced in my own project accidentally...

So I take it this removal of the `root` section prevents something similar?

montchr (approved):
LGTM, with one suggestion

cells/std/nixago/editorconfig.nix
montchr: of course, all that said, i don't recall ever having noticed a difference with having `root = true` or `root = false` or omitting it entirely 🤷‍♀️

cells/std/nixago/editorconfig.nix
blaggacao: Yeah, this is just the way that nix' `toTOML` is implemented.

PR #195, opened 2022-11-28
wip: initial spin off of paisano

blaggacao:


blaggacao:
This is paused until we have clarity over the new improved action interface that is coming out of the discussions during implementation of https://github.com/divnix/std-action

PR #200, opened 2022-12-02
feat: add a `check` subcommand to validate the repo

### LINKED ISSUE #198: Add a `check` subcommand to non-interactively
It would be great to add a  `std check` subcommand that could verify that the repo evaluates. 

I've noticed that running the std  can break when nix code changes, and you don't find out until you try to run the TUI. `nix flake check` can take a while, and give error messages that are much less helpful than the ones provided by `std` when it fails. It would be great to have a non-interactive way of making sure the repo is correct. This command would be useful as something that could be run in CI.

Later on this could be extended with flags to enable something like `std check --comply-with set-of-rules.nix`. i.e. some sort of "doctor" command.

I could work on a PR for this.

tshaynik:
Closes #198 

As of now, its implementation is similar to re-cache, with a few important differences:
- Stderr is printed
- It returns a nonzero exit code if it fails to evaluate
- It prints a success message if evaluation succeeds

### DIFF
--- a/cells/std/cli/cli.go
+++ b/cells/std/cli/cli.go
@@ -76,6 +76,27 @@ The TUI does this automatically, but the command completion needs manual initial
 		return nil
 	},
 }
+var checkCmd = &cobra.Command{
+	Use:   "check",
+	Short: "Validate the repository.",
+	Long: `Validates that the repository conforms to Standard.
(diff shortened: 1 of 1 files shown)


blaggacao:
LGTM, let it loose!

PR #202, opened 2022-12-17
fix: action invocation on the actual git repo

blaggacao:
using the flake output derivation effectively invokes nix
commands on a source copy that isn't a git repository anymore
hence `self.rev` does not exists. That means that any intentional
use of `self.rev` was sabotaged by the use of std actions.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -41,7 +41,22 @@
     arion.follows = "blank";
   };
   outputs = inputs: let
-    blockTypes = import ./src/blocktypes.nix {inherit (inputs) nixpkgs;};
+    blockTypes = import ./src/blocktypes.nix {
+      inherit deSystemize;
+      nixpkgs = builtins.mapAttrs (system: pkgs:
+        pkgs
+        // {
+          writeShellScriptWithPrjRoot = name: content:
+            pkgs.writeShellScript name ''
+              if test -z "$PRJ_ROOT"; then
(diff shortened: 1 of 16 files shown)


jmgilman:
LGTM. As noted in matrix, this will break `std-action`. It needs to be updated to checkout a local copy of the repository in the run step.

nrdxp:
That doesn't really make sense. `self.rev` should exist on any reference that is a clean git remote.

blaggacao:
> writable copy

That is indeed a different promise than being a mere git repo. I agree. :+1:

jmgilman:
Can we compromise on a short-term fix? This is a blocker for me downstream :)

blaggacao:
@jmgilman can you test if this works?

If confirmed, then I'll merge with a deprecation instrumentation.

jmgilman:
I'll test this tomorrow and report back.

nrdxp:
> Can we compromise on a short-term fix? This is a blocker for me downstream :)

I think it'd be pretty quick to just fix the action in std proper. I'll look into it here after my morning meetings.

nrdxp:
> I'll test this tomorrow and report back.



> Can we compromise on a short-term fix? This is a blocker for me downstream :)

I don't think a proper fix would take long at all. I'll give it a shot right now, as I'd rather not have to backtrack

blaggacao:
closing in favor of #206

Issue #218, opened 2022-12-23
Add nixago presetr for std-action


blaggacao:

blaggacao:
Autogenerating this doesn't seem fit for purpose. We can't get the magic all the way and when flexibility is needed, this is a hinderance.

PR #219, opened 2022-12-23
feat: add initial std-gh-action nixago instrumentation

blaggacao:
closes: #218

### DIFF
--- a/cells/_automation/nixago.nix
+++ b/cells/_automation/nixago.nix
@@ -69,4 +69,19 @@ in {
       ];
     };
   };
+  std-gh-action = presets.nixago.std-gh-action {
+    configData = {
+      on.push.branches = ["main"];
+      targets = [
+        {
+          block = "cli";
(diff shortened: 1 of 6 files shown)


cells/_automation/nixago.nix
```diff
@@ -69,4 +69,19 @@ in {
...
+      on.push.branches = ["main"];
+      targets = [
+        {
+          block = "cli";
+          action = "build";
+        }
+        {
+          block = "packages";
+          action = "build";
+        }
```
blaggacao, line 83: Unfortunately, I don't see a way how we could scrape this from `__std.ci`. Any ideas?

blaggacao:
Now, how do we enable this for mac, too?

blaggacao:
Autogenerating this doesn't seem fit for purpose. We can't get the magic all the way and when flexibility is needed, this is a hinderance.

Issue #224, opened 2022-12-24
Seamlessly integrate with `flake-parts`


blaggacao:
Interoperability is key.

Issue #211, opened 2022-12-22
Add quick-start templates


blaggacao:
Scenarios to cover:
- SDLC
- Maybe NixOS (via `divnix/hive`)?

Maybe those should reside in a different repository for clarity.

blaggacao:
- [minimal](https://github.com/divnix/std/tree/main/cells/presets/templates/minimal)

PR #226, opened 2022-12-24
feat: impl flake parts module

blaggacao:
closes: #224

@roberth @dermetfan can you have a look at this?

### DIFF
--- a/src/flakeModule.nix
+++ b/src/flakeModule.nix
@@ -1,0 +1,180 @@
+{
+  validators,
+  grow,
+  pick,
+  winnow,
+  harvest,
+}: {
+  inputs,
+  config,
+  options,
+  lib,
+  ...
+}: let
+  inherit
+    (builtins)
+    all
+    concatMap
+    head
+    isList
+    length
+    mapAttrs
+    tail
+    zipAttrsWith
+    ;
+  inherit
+    (lib)
+    getValues
+    mkOption
+    mkOptionType
+    showDefs
+    showOption
+    types
+    literalExpression
+    ;
+
+  # ensures we always end up with a list of lists
+  # if more than one definition
+  mergeHarvesterOption = loc: defs: let
+    list = getValues defs;
+  in
+    if length list == 1
+    then head list
+    else if all isList list
+    then
+      concatMap (l:
+        if isList l
+        then l
+        else [l])
+      list
+    else throw "Cannot merge definitions of `${showOption loc}', as at least one merge candidate is not a list of strings or a list of list of strings. Definition values:${showDefs defs}";
+
+  harvesterType = mkOptionType {
+    name = "harvester";
+    description = "harvest path(s) as a list of strings or a list of lists of strings";
+    descriptionClass = "noun";
+    merge = mergeHarvesterOption;
+    emptyValue = {value = [];};
+    inherit
+      (types.either
+        (types.listOf types.nonEmptyStr)
+        (types.listOf (types.listOf types.nonEmptyStr)))
+      check
+      ;
+  };
+
+  cellBlocksType = mkOptionType {
+    name = "cellBlocks";
+    description = "list of cell block";
+    descriptionClass = "noun";
+    inherit (validators.CellBlocks) check;
+  };
+
+  opt = options.std;
+  cfg = config.std;
+in {
+  _file = ./flakeModule.nix;
+  options = {
+    std = {
+      grow = mkOption {
+        description = ''
+          Orderly 'grow' a project from Standard Cells & Cell Blocks.
+           - Find the [glossary here](https://std.divnix.com/glossary.html).
+           - Find a good [walk-through here](https://jmgilman.github.io/std-book/).
+           - And the general [documentation here](https://std.divnix.com/index.html).
+        '';
+        type = types.submodule {
+          options = {
+            cellsFrom = mkOption {
+              description = "Where Standard discovers Cells from.";
+              type = types.path;
+              example = literalExpression "./nix";
+            };
+            cellBlocks = mkOption {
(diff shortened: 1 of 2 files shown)


blaggacao:
noticed an issue, will fix tomorrow

src/flakeModule.nix
```diff
@@ -0,0 +1,138 @@
...
+      winnow = mkOption {
+        description = "Winnow Standard outputs. Like 'harvest', but with a filter predicate 'winnowIf'.";
+        type = types.attrsOf harvesterType;
+      };
+      harvest = mkOption {
+        description = "Harvest Standard outputs into a Nix-CLI-compatible form (a.k.a. the 'official' flake schema).";
+        type = types.attrsOf harvesterType;
+      };
+    };
+  };
```
roberth, line 114: I'd recommend to add examples to all of the options, and perhaps expand the documentation a bit, possibly by linking to existing documentation. `description` is markdown, ahead of the Nixpkgs/NixOS switch to markdown.

src/flakeModule.nix
```diff
@@ -0,0 +1,138 @@
...
+    in
+      lib.foldl' lib.recursiveUpdate {} (
+        [grown]
+        ++ (lib.optionals opt.pick.isDefined [picked])
+        ++ (lib.optionals (opt.winnow.isDefined && opt.winnowIf.isDefined) [winnowed])
+        ++ (lib.optionals opt.harvest.isDefined [harvested])
+      );
+
+    # exposes the raw scheme of the std layout inside flake-parts
+    perInput = system: flake: lib.optionalAttrs (lib.hasAttrs system flake) flake.${system};
```
roberth, line 136: I'd recommend namespacing this, to avoid collisions. `perInput` is applied to _all_ inputs to form `inputs'`, so setting arbitrary attributes could lead to nasty interactions. I'll improve the `perInputs` description.

```suggestion
    perInput = system: flake: { std = flake.${system} or {}; };
```

src/flakeModule.nix
roberth: An complete example in the form of a flake template would also be helpful.

blaggacao: > An complete example in the form of a flake template would also be helpful.

That's in planning as part of https://github.com/divnix/std/issues/211

I don't want to treat it in an ad-hoc way...

src/flakeModule.nix
```diff
@@ -0,0 +1,167 @@
...
+    if length list == 1
+    then head list
+    else if all isList list
+    then
+      concatMap (l:
+        if isList l
+        then l
+        else [l])
+      list
+    else throw "Cannot merge definitions of `${showOption loc}'. Definition values:${showDefs defs}";
```
roberth, line 50:
```suggestion
    else throw "Cannot merge definitions of `${showOption loc}', as at least one list item is not a list of strings. Definition values:${showDefs defs}";
```

Is this correct?

src/flakeModule.nix
```diff
@@ -0,0 +1,167 @@
...
+
+  harvesterType = mkOptionType {
+    description = "harvest path as list (or list of lists) of strings";
```
roberth, line 53:
```suggestion
    description = "harvest path(s) as list (or list of lists) of strings";
```

src/flakeModule.nix
```diff
@@ -0,0 +1,167 @@
...
+              '';
+            };
+          };
+        };
+      };
+      pick = mkOption {
+        description = "Pick Standard outputs. Like 'harvest' but remove the system for outputs that are system agnostic.";
+        type = types.attrsOf harvesterType;
+        example = literalExpression ''
+          { lib = [ "utils" "library" ];
```
roberth, line 111:
```suggestion
          { lib = [ "utils" "library" ]; }
```

src/flakeModule.nix
roberth: Judging from the :+1: I think you forgot this one.

src/flakeModule.nix
```diff
@@ -0,0 +1,167 @@
...
+    else if all isList list
+    then
+      concatMap (l:
+        if isList l
+        then l
+        else [l])
+      list
+    else throw "Cannot merge definitions of `${showOption loc}'. Definition values:${showDefs defs}";
+
+  harvesterType = mkOptionType {
```
roberth, line 52:
```suggestion
  harvesterType = mkOptionType {
    name = "harvester";
```

This name is internal to the module system, but required for the documentation to eval. Odd.

src/flakeModule.nix
```diff
@@ -0,0 +1,167 @@
...
+    emptyValue = {value = [];};
+    inherit
+      (types.either
+        (types.listOf types.nonEmptyStr)
+        (types.listOf (types.listOf types.nonEmptyStr)))
+      check
+      ;
+  };
+
+  cellBlocksType = mkOptionType {
```
roberth, line 65:
```suggestion
  cellBlocksType = mkOptionType {
    name = "cellBlocks";
```

same

roberth (commented):
A link to your main documentation would be nice, but you can add that later if you prefer.

After applying the internal type `name` suggestions, you can preview the docs with

```console
nix build github:hercules-ci/flake.parts-website/add-std --override-input std .
```

If you agree that I publish the generated docs on https://flake.parts, I'll merge https://github.com/hercules-ci/flake.parts-website/pull/82 and they'll be auto updated.

src/flakeModule.nix
blaggacao: Not entirely, I'll reformulate.

blaggacao:
> If you agree that I publish

Awesome! Please do!

PR #229, opened 2023-01-03
feat(ci): add proviso for build

nrdxp:
A proviso is an optional condition for a given action which std-action can use to determine if a job actually needs to run in a given CI iteration, saving work if not.

Proviso should be free from context unless there is absolutely no workaround, and they should be idempotent, at least in a pragmantic sense.

linked with divnix/std-action#28

~~Will take out of draft after I implement a proviso for OCI image publishing and nail down the spec.~~ done

### DIFF
--- a/src/blocktypes.nix
+++ b/src/blocktypes.nix
@@ -2,7 +2,10 @@
   mkCommand = system: type: args:
     args
     // {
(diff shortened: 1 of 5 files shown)


blaggacao (approved):
Nice!

PR #230, opened 2023-01-06
fix: allow overriding the targetDrv

nrdxp:
This value is rightfully coupled to what is actually placed in the action script.


blaggacao:
&rarr; paisano

PR #236, opened 2023-01-09
fix: violently nudge the use of an up-to-date nix version

blaggacao:
it's a fast-paced world now-a-days...

We had a user that hit an unbound feedback loop strolling into an UX abyss on `nix 2.5`...

- 2.7 has `.default` over `defaultX` semantics
- 2.9, iirc, has the vital `__funtor` fix


blaggacao:
&rarr; paisano

PR #239, opened 2023-01-10
Initiate Standard Action

nrdxp:
Also includes Mac builds so that our cache is popullated for Mac users.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,20 @@
+name: Standard CI
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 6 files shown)


blaggacao (approved):
Vacano!

blaggacao:
@nrdxp would you want to take over https://github.com/divnix/std/pull/219 from me? (Since you figured out how to do macos) :smile:

nrdxp:
Well I think you are still more familiar with Nixago. In order to build for more systems in general you just need to use a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows).

PR #241, opened 2023-01-14
wip: implement mdbook-std-reference preprocessor

blaggacao:

PR #242, opened 2023-01-14
refactor(cli): make readme an optional field and don't system space init

blaggacao:
This is a preliminary refactoring for #241


blaggacao:
&rarr; paisano

PR #247, opened 2023-01-20
imp: enrich the microvm command set

GTrunSec:
```
std //guangtao/microvms/dev:run               
char device redirected to /dev/pts/10 (label con0)
checking /dev/vdb...
[fsck.ext4 (1) -- /mnt-root/var] fsck.ext4 -a /dev/vdb
/dev/vdb: recovering journal
/dev/vdb: clean, 60/131072 files, 30342/524288 blocks

<<< NixOS Stage 2 >>>

running activation script...
reviving group 'dhcpcd' with GID 999
reviving group 'nscd' with GID 998
```

### DIFF
--- a/src/blocktypes/microvms.nix
+++ b/src/blocktypes/microvms.nix
@@ -7,6 +7,8 @@
   Use the Microvms Blocktype for Microvm.nix - https://github.com/astro/microvm.nix
 
   Available actions:
+    - run
+    - console
     - microvm
   */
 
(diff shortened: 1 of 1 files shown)


nrdxp (approved):
Looks fine to me. I haven't messed with microvms more than a few times so I can't recall, are there other commands besides just these? If so, why not just enumerate them?

GTrunSec:
Other subcommands of microvm are not useful so far. I have already discussed this problem with Astro, https://github.com/Mic92/vmsh would be the solution. But mic92 doesn't have time to update vmsh to the recent kernel.

PR #259, opened 2023-02-15
[WIP] Use `contracts.yants` as runtime types provider

yvan-sraka:
This is an attempt to test my ongoing work https://github.com/yvan-sraka/contracts/pull/1 against the CI of a real-world project like `std` :)


yvan-sraka:
CI passed spontaneously 🤯, still have a few things on my To-Do before merging:

- [ ] For better logging / tracing context, `std` been using a `__functor ` / a forked version of `yants` from [here](https://github.com/divnix/yants/blob/d7253ea436060f598c5d0ef3b22594fd28e7eab2/flake.nix#L367-L390), a similar DevX should be offered by `contracts` ;

- [ ] Remove a bit of a glue, since `contracts` can reuse types defined `nixpkgs.lib.type` and use `contracts` types as NixOS option:
https://github.com/divnix/std/blob/f437f107feafc2984bb2b2f98fe8b13022021746/src/flakeModule.nix#L70

PR #262, opened 2023-02-17
feat: polish the mkOCI interface further by exposing oft-used 'config'

blaggacao:
When images need more configuration than the default exposed options,
in most cases, it is concerned with n2c's 'config' iface.

To honor that reality, let's expose it.

While this adds two similar interfaces (options & config), in practice,
when people copy over examples they would only see:

```nix
{
  config = /* snip */;
}
```

That is less confusing than the following.
This was also the semantic hickup that prompted this commit:
```nix
{
  options.config = /* snip */;
}
```


jmgilman:
LGTM.

PR #263, opened 2023-02-18
fix: mkOperableScript needs args list

nrdxp:
An atribute set makes single flag arguments awkward to declare, and no specific order is guaranteed, meaning that programs that require args in a particular order will fail.

In order to use the style which is most natural for the given program and to keep the resulting script readable, users can group key=value style arguments in a true attribute set, or several arguments of the same class in a list; i.e:
```nix
mkOperableScript {
  args = [
    "--single-flag"
    {"--port" = 80;}
    ["path/a" "path/b" "path/c" ["path/with" "spaces"]]
  ];
  # ...
}
```
results in:
```bash
args+=("--single-flag")
args+=("--port" "80")
args+=("path/a" "path/b" "path/c" "path/with spaces")
```

For the sake of backward compatibility, the previous API is retained in the event that an attribute set is passed instead of a list.

### DIFF
--- a/cells/lib/ops/mkOperableScript.nix
+++ b/cells/lib/ops/mkOperableScript.nix
@@ -2,8 +2,7 @@
   inputs,
   cell,
 }: let
-  inherit (inputs) nixpkgs std;
-  l = nixpkgs.lib // builtins;
+  l = inputs.nixpkgs.lib // builtins;
 in
   /*
   Makes a simple wrapper for executing an operable's package with args.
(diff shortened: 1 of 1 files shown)


cells/lib/ops/mkOperableScript.nix
```diff
@@ -18,13 +17,34 @@ in
...
+        prefix ? "",
+        suffix ? "",
+      }:
+        builtins.concatStringsSep "\n" (
+          l.mapAttrsToList (name: value: ''${prefix}"${name}" "${toString value}"${suffix}'') set
+        );
+    in
+      if builtins.isAttrs args
+      then
+        attrsToStr {
```
blaggacao, line 36: Can you incorporate a deprecation message in `deprecations.nix` and invoke it for this code path?

blaggacao: ok, not needed, it's still a hoisting simplification of the singleton `[{}]`.

blaggacao (approved):
LGTM

PR #266, opened 2023-02-22
feat: allow flakeModule usage without growing anything

roberth:
This allows the module to be used in a situation where the caller consumes `inputs'.foo.cells` but does not use the std options to define its own outputs.


blaggacao:
Thank you, that's a nice interop addition!

PR #267, opened 2023-02-23
feat: acquire ownership over container actions from n2c

blaggacao:
This lays the groundwork for more control over how we implement them.

Users can now set an arbitrary number of tags for an image and upload all the given tags. To specify tags, set `meta.tags` to a list of your desire tags in the arguments to `mkOCI` or `mkStandardOCI`. 

The first tag is special in that it is used as the tag checked in the proviso for existence. That way, one can use a Nix input hash as the first tag to only upload when the derivation inside the image actually changes, while still tagging the images that are uploaded with useful tags like the git revision.

If no tags are set, the default upstream tagging mechanism of nix2container is used.

### DIFF
--- a/src/blocktypes/containers.nix
+++ b/src/blocktypes/containers.nix
@@ -26,7 +26,11 @@
     }: let
       inherit (n2c.packages.${currentSystem}) skopeo-nix2container;
       copy-to = "${skopeo-nix2container}/bin/skopeo --insecure-policy copy nix:${target}";
-      imageRef = target.imageRefUnsafe or (builtins.unsafeDiscardStringContext "${target.imageName}:${target.imageTag}");
+      img = builtins.unsafeDiscardStringContext target.imageName;
+      tags = let
+        tags' = target.meta.tags or [(builtins.unsafeDiscardStringContext target.imageTag)];
+      in
+        builtins.toFile "${target.name}-tags" (builtins.unsafeDiscardStringContext (builtins.concatStringsSep "\n" tags'));
     in [
       (sharedActions.build currentSystem target)
       (mkCommand currentSystem {
@@ -41,8 +45,9 @@
         name = "publish";
         description = "copy the image to its remote registry";
         command = ''
-          # docker://${imageRef}
-          ${copy-to} docker://${imageRef} $@
+          for tag in $(<${tags}); do
+            ${copy-to} "docker://${img}:$tag" "$@"
+          done
         '';
         proviso =
           l.toFile "container-proviso"
@@ -57,8 +62,7 @@
 
             function get_images () {
               command nix show-derivation $@ \
-              | command jq -r '.[].env.text' \
-              | command grep -o 'docker://\S*'
+              | command jq -r '.[].env.__provisory'
             }
 
             drvs="$(command jq -r '.actionDrv | select(. != "null")' <<< "''${input[@]}")"
@@ -90,18 +94,28 @@
             output=$(command jq -cs '. += $p' --argjson p "$output" <<< "''${filtered[@]}")
             }
           '';
+
+        provisory = target.meta.provisory or "${img}:${target.imageTag}";
       })
       (mkCommand currentSystem {
         name = "load";
         description = "load image to the local docker daemon";
         command = ''
+          copy_tags() {
+            local storage=$1
+            shift
+
+            for tag in $(<${tags}); do
+              ${copy-to} "$storage:${img}:$tag" "$@"
+            done
+          }
           if command -v podman &> /dev/null; then
              ixecontainerho "Podman detected: copy to local podman"
-            ${copy-to} containers-storage:${imageRef} $@
+             copy_tags containers-storage "$@"
           fi
           if command -v docker &> /dev/null; then
              echo "Docker detected: copy to local docker"
-            ${copy-to} docker-daemon:${imageRef} $@
+             copy_tags docker-daemon "$@"
           fi
         '';
       })
--- a/src/blocktypes/containers.nix
+++ b/src/blocktypes/containers.nix
@@ -10,9 +10,8 @@
 
   Available actions:
     - print-image
-    - copy-to-registry
-    - copy-to-podman
-    - copy-to-docker
+    - publish
+    - load
   */
   containers = name: {
     __functor = import ./__functor.nix;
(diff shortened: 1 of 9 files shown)


src/mkCommand.nix
```diff
@@ -1,5 +1,13 @@
 {nixpkgs}: let
-  writeShellScript = system: nixpkgs.legacyPackages.${system}.writeShellScript;
-  mkCommand = system: args: args // {command = writeShellScript system "${args.name}" args.command;};
+  mkCommand = system: args: let
+    inherit (nixpkgs.legacyPackages.${system}) pkgs;
+  in
+    args
+    // {
+      command = (pkgs.writeShellScript "${args.name}" args.command).overrideAttrs (_:
```
blaggacao, line 7: https://github.com/nlewo/nix2container/pull/65 was merged :smile:

blaggacao (commented):
Upstream merged your amendment.

src/mkCommand.nix
nrdxp: Nice I updated the input and remove this logic

nrdxp:
I've unit tested the proviso and the new publish action so this should be ready to merge. In the future we might want to come back and setup a mock registry to automate unit testing for proviso.

src/blocktypes/containers.nix
```diff
@@ -35,79 +67,46 @@
...
-        name = "copy-to-registry";
-        description = "copy the image to its remote registry";
+        name = "load";
+        description = "load image to the local docker daemon";
         command = ''
-          ${target.copyToRegistry}/bin/copy-to-registry
+          ${copyTags}
+
+          if command -v podman &> /dev/null; then
+             ixecontainerho "Podman detected: copy to local podman"
```
blaggacao, line 87: `ixecontainerho`? :smile:

src/blocktypes/containers.nix
```diff
@@ -22,7 +23,38 @@
...
+      img = builtins.unsafeDiscardStringContext target.imageName;
+      tags = target.meta.tags or [(builtins.unsafeDiscardStringContext target.imageTag)];
+      copyTags = let
+        skopeo = "skopeo --insecure-policy";
+        tags' =
+          builtins.toFile "${target.name}-tags.json" (builtins.unsafeDiscardStringContext (builtins.concatStringsSep "\n" tags));
+      in ''
+        export PATH=${skopeo-nix2container}/bin:$PATH
+
+        copy_tags() {
```
blaggacao, line 37: Maybe `copy_tags` is a misnomer, since this effectivly loads/publishes an entire image (if need be).

src/blocktypes/containers.nix
```diff
@@ -22,7 +23,38 @@
...
+          local name tags source img
+          source=nix:${target}
+          tags=${tags'}
+          name=$1
```
blaggacao, line 41: Since it includes the schema-ish (`docker://` / `docker-daemon` / etc), `name` might not be the right name. Maybe `sourceURI` & `targetURI`, as a better alternative?

src/blocktypes/proviso/publish.sh
```diff
@@ -0,0 +1,29 @@
...
+  local -n input=$1
+  local -n output=$2
+
+  local missing tag
+  local -a images filtered
+
+  missing=$(mktemp)
+  trap 'rm -f $missing' RETURN
+
+  mapfile -t images < <(jq -r '.meta.tags[0]|select(.!=null)' <<<"${input[@]}")
```
blaggacao, line 13: Is that really plural? I don't fully understand the jq foo but `tags[0]` looks like a `head` operation that would result with a singleton.

src/blocktypes/proviso/publish.sh
```diff
@@ -0,0 +1,29 @@
...
+
+  echo "${images[@]}" |
+    command xargs -n 1 -P 0 "$SKOPEO" "$missing"
```
blaggacao, line 16: By `$SKOPEO` name I was enticed to beleive this is what skopeo normally does in this context, but no, it's more like `$SKOPEO_INSPECT`. Reading it this way, this line starts to make more sense.

Can we rename?

src/mkCommand.nix
```diff
@@ -1,5 +1,10 @@
 {nixpkgs}: let
-  writeShellScript = currentSystem: nixpkgs.legacyPackages.${currentSystem}.writeShellScript;
-  mkCommand = currentSystem: args: args // {command = writeShellScript currentSystem "${args.name}" args.command;};
+  mkCommand = system: args: let
```
blaggacao, line 2: This is now notably `currentSystem` and should keep that name. :smile:

src/blocktypes/containers.nix
```diff
@@ -22,7 +23,38 @@
       fragment,
       fragmentRelPath,
       target,
-    }: [
+    }: let
```
blaggacao, line 26: I think we overlooked:

`print-image` action (in case somebody uses it). Could be plural, now...

And also the docstring needs update.

blaggacao (commented):
This looks fantastic! There are some oversights and a couple of things where the code readability could be a bit polished (from my perspective).

src/blocktypes/containers.nix
nrdxp: you wrote that in the first commit here so I thought it was intentional. Typo maybe?

src/blocktypes/containers.nix
nrdxp: right we are copying tags to the image registry. I used the same sematic as copy_to

src/blocktypes/containers.nix
blaggacao: Yeah, oops :-)

src/blocktypes/proviso/publish.sh
nrdxp: yes we only want to check the first tag, that's explicit (see update PR description).

src/blocktypes/proviso/publish.sh
nrdxp: sure

src/mkCommand.nix
nrdxp: ah yeah, mistake during rebase

src/blocktypes/containers.nix
blaggacao: Maybe just `copy` so that it can mean both? I don't think about a "payload", when I hear the word "tag", which is true for n+1, indeed. But not for n.

src/blocktypes/proviso/publish.sh
blaggacao: I mean: is it still `images` (plural), though?

src/blocktypes/containers.nix
nrdxp: should we just stick to printing the first tag or print them all in a loop?

blaggacao: I would almost vote for "all in a loop".

It would be a breaking change, but then, this action's purpose is utterly informational. So all the tags seems like relevant information.

People are unlikely to use that for scripting (or so I hope!!). That's what the other actions are for, after all.

src/blocktypes/proviso/publish.sh
nrdxp: yeah because we take the first tag over all the input images. Remember, the proviso runs over all the inputs not just a single one.

blaggacao: Ah yeah! Click. :smile:

src/blocktypes/containers.nix
nrdxp: The list of tags in the meta would be more suited for scripting purposes anyway

src/blocktypes/containers.nix
nrdxp: just went the explicit route to make it as clear as possible what's happening

src/blocktypes/containers.nix
```diff
@@ -22,92 +23,88 @@
...
-            done
-
-            output=$(command jq -cs '. += $p' --argjson p "$output" <<< "''${filtered[@]}")
-            }
-          '';
+        proviso = l.toFile "container-proviso" ''
+          SKOPEO_INSPECT=${./proviso/skopeo-inspect.sh}
+          ${builtins.readFile ./proviso/publish.sh}
+        '';
+        meta.tags = map (tag: "${img}:${tag}") tags;
```
blaggacao, line 77: Maybe we also should name this something else? As it's not `tags`, anymore? It's probably correctly `images`?

blaggacao (commented):
One additional clarification may help to parse the code.

Can you also update the action docstring? It's shown verbatim in the docs. I had forgotten to do that in my commits.

src/blocktypes/containers.nix
nrdxp: yeah I suppose that makes sense

blaggacao: That would also clarify `tags[0]` to `images[0]` in the proviso.

src/blocktypes/proviso/publish.sh
```diff
@@ -0,0 +1,29 @@
...
+  missing=$(mktemp)
+  trap 'rm -f $missing' RETURN
+
+  mapfile -t images < <(jq -r '.meta.images[0]|select(.!=null)' <<<"${input[@]}")
+
+  echo "${images[@]}" |
+    command xargs -n 1 -P 0 "$SKOPEO_INSPECT" "$missing"
+
+  for in in "${input[@]}"; do
+    tag=$(jq -r '.meta.images[0]' <<<"$in")
```
blaggacao, line 19: Now we have an `image`, here, too. :smile:

blaggacao (commented):
Last thing? :smile:

src/blocktypes/proviso/publish.sh
nrdxp: oh whoops

PR #271, opened 2023-02-26
fix: ensure old image tag api is consistent

nrdxp:
We don't want users to hit a situation where  they reference one of the legacy n2c APIs such as `imageTag` and get the wrong value.

Therefore, disallow `tag` & `meta.tags` from being set together, always set the first value in `meta.tags` to the legacy `tag` arg of nix2container's `buildImage` function, and warn the user of the deprecated interface when accessed.

### DIFF
--- a/cells/lib/ops/mkOCI.nix
+++ b/cells/lib/ops/mkOCI.nix
@@ -25,10 +25,13 @@ in
   Returns:
   An OCI container image (created with nix2container).
   */
-  {
+  args @ {
     name,
     entrypoint,
-    tag ? "",
+    tag ?
+      if meta ? tags && (l.length meta.tags) > 0
+      then l.head meta.tags
+      else "",
     setup ? [],
     layers ? [],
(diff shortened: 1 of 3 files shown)


cells/lib/ops/mkOCI.nix
```diff
@@ -81,6 +84,8 @@ in
         # Setup tasks can include permissions via the passthru.perms attribute
         perms = l.flatten ((l.map (s: l.optionalAttrs (s ? passthru && s.passthru ? perms) s.passthru.perms)) setup) ++ perms;
       }
-      // l.optionalAttrs (tag != "") {inherit tag;};
+      // l.throwIf (args ? tag && meta ? tags)
+      "mkStandardOCI: use of `tag` and `meta.tags` arguments are not supported together. Remove the former."
```
blaggacao, line 88: `mkOCI`?

blaggacao (approved):
Just a minor copy past error...

PR #273, opened 2023-02-28
feat: add /bin/sh into container by default

nrdxp:
In order for BASH's posix emulation to work correctly, the basename of `$0` must be exactly `sh`, so we set `exec -a "$0"` to allow this to work properly.

In addition, use `buildEnv` for all `setup` links since this helps to automatically resolve package priority conflicts where possible.


cells/lib/ops/mkStandardOCI.nix
```diff
@@ -41,7 +41,10 @@ in
     meta ? {},
   }: let
     # Link useful paths into the container.
-    runtimeEntryLink = "ln -s ${l.getExe operable.passthru.runtime} $out/bin/runtime";
+    runtimeEntryLink = ''
+      ln -s ${l.getExe operable.passthru.runtime} $out/bin/sh
```
jmgilman, line 45: It's been awhile since I've visited this. Why are linking the runtime to /bin/sh?

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -62,7 +65,7 @@ in
       else operable;
 
     setupLinks = cell.ops.mkSetup "links" [] ''
-      mkdir -p $out/bin
+      mkdir -p $out/{bin,tmp}
```
jmgilman, line 68: Should this be larger discussion about what a default container should look like? I have more ideas we could add, but I opted to leave it up to the end user to implement.

cells/lib/ops/mkOCI.nix
```diff
@@ -70,7 +70,10 @@ in
           ++ layers;
 
         maxLayers = 25;
-        copyToRoot = [setupLinks] ++ setup;
+        copyToRoot = nixpkgs.buildEnv {
```
jmgilman, line 73: What benefit is this bringing?

nrdxp: without this, any to files in `copyToRoot` will cause a conflict, even if its just a directory with the same name but different contents. `buildEnv` resolves these conflicts based on pakage priority, so unless two packages have the same exact priority, they won't conflict.

Also there is no conflict on directories, so they will just be merged together. Not only is this change not possible without this (for whatever reason it triggered a conflict), but it will make the root file system easier for the user to extend as they need to.

cells/lib/ops/mkStandardOCI.nix
nrdxp: potentially, but I feel like `/tmp` is fairly harmless on its own

cells/lib/ops/mkStandardOCI.nix
nrdxp: I could just link the `runtimeShell` directly, but the runtime has the nice property of setting the path to the runtime inputs, so if a program depends on `/bin/sh` and that invocation calls out to some other program that already exists in the `runtimeInputs` it would suceed. If we just linked directly to `nixpkgs.runtimeShell` it would fail since the PATH would basically be empty.

jmgilman: Ah, that makes sense.

cells/lib/ops/mkStandardOCI.nix
jmgilman: Yeah, I suppose it just feels out of place, given the intent of the commit message. I guess I was suggesting maybe parsing it out into a separate discussion that addresses the more significant issue of having a more helpful "base image" generated by the function.

cells/lib/ops/mkOCI.nix
jmgilman: I'm surprised I haven't run into this before 🤷 SGTM.

cells/lib/ops/mkStandardOCI.nix
GTrunSec: > more significant issue of having a more helpful "base image" generated by the function.

Maybe we use `preset=false` to disable some configurations and keep a primitive image.

nrdxp: I can just remove the commit that adds /tmp for now and do a separate change-set for that. Good idea

nrdxp: > Maybe we can use `preset=false or pure` to disable some configurations and keep a primitive image.

This or something similar is likely what we should do in the end. I opened #274 as a draft for now to hash that out further.

blaggacao:
LGTM

PR #274, opened 2023-03-01
fix: create /tmp by default for containers

nrdxp:
Draft for a potentially richer, optional layer on top of the base image.

brainstorm:
- [ ] existing directories: /tmp, /srv, etc
- [ ] basic /etc: /etc/group, /etc/passwd, etc
- [ ] more powerful debug tools `rg`, `fd` etc
- [ ] potential base image (alpine?)
- [ ] ?

### DIFF
--- a/cells/lib/ops/mkStandardOCI.nix
+++ b/cells/lib/ops/mkStandardOCI.nix
@@ -2,7 +2,7 @@
   inputs,
   cell,
 }: let
-  inherit (inputs) nixpkgs std;
+  inherit (inputs) nixpkgs;
   l = nixpkgs.lib // builtins;
   n2c = inputs.n2c.packages.nix2container;
 in
@@ -62,7 +62,7 @@ in
       else operable;
 
     setupLinks = cell.ops.mkSetup "links" [] ''
-      mkdir -p $out/bin
+      mkdir -p $out/{bin,tmp}
       ${runtimeEntryLink}
       ${debugEntryLink}
       ${livenessLink}


blaggacao:
All these things make it "easier", by fulfilling generally accepted, but largely undocumented ecosystem expectations.

One of the beauties of Nix is that it educates users to make them explicit.

And being explicit about an operable's other-than-env-variable runtime contract is valuable information, in my opinion.

So generally, I'm erring on the side of conservatism about this.

I'm also conservative about the idea of being "optional". Imagine a fridge with 20 knobs and switches. I already "freak out" when I see a microwave I'm not familiar with. There is an underlying truth to this, even in engineering. So also im that regard, I'd want the consideration to be extremely cognizant about this trade-off.

blaggacao:
We may explore to provide a set of setup packages, e.g. for the first two ideas.

But I'm definitely out on the third idea since "more advanced" here also means "the ones _we_ use" and this repo is not meant to be out private backyard (like bitte was). 😊

jmgilman:
> One of the beauties of Nix is that it educates users to make them explicit.

It's a tradeoff, though. It directly hurts adoption since the most common pushback I see is: "I don't want to learn Nix!" Even more, the whole `nix2container` is so foreign to anyone not baptized into the Nix Docker oddities that it's likely more frustrating having a "very blank" image as a start. 

I'm not sold one way or the other yet. Just playing devil's advocate on the other side of the fence.

jmgilman:
I'll paste some code from another project which I've been using to setup some non-production images:

```nix
  containerCommon = {
    uid = "1000";
    gid = "1000";

    setup = let
      setupUser = std.lib.ops.mkUser {
        user = "user";
        group = "user";
        uid = "1000";
        gid = "1000";
        shell = nixpkgs.runtimeShell;
        withHome = true;
        withRoot = true;
      };

      setupEnv =
        std.lib.ops.mkSetup "container"
        [
          {
            regex = "/tmp";
            mode = "0777";
          }
        ]
        ''
          # Setup tmp folder
          mkdir -p $out/tmp

          # Enable nix flakes
          mkdir -p $out/etc
          echo "sandbox = false" > $out/etc/nix.conf
          echo "experimental-features = nix-command flakes" >> $out/etc/nix.conf

          # Put local profile in path
          echo 'export PATH="$HOME/.nix-profile/bin:/nix/var/nix/profiles/default/bin:$PATH"' >> $out/etc/bashrc

          # Setup certificates
          cp -r ${nixpkgs.cacert}/* $out
        '';
    in [setupEnv setupUser];

    options = {
      initializeNixDatabase = true;
      nixUid = 1000;
      nixGid = 1000;

      config = {
        Env = [
          # Required by many tools
          "HOME=/home/user"
          # Nix related environment variables
          "NIX_CONF_DIR=/etc"
          "NIX_PAGER=cat"
          # This file is created when nixpkgs.cacert is copied to the root
          "NIX_SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt"
          # Nix expects a user to be set
          "USER=user"
          # Include <nixpkgs> to support installing additional packages
          "NIX_PATH=nixpkgs=${nixpkgs.path}"
        ];
      };
    };
  };
```

nrdxp:
> I'm not sold one way or the other yet. Just playing devil's advocate on the other side of the fence.

It's incredibly valuable to challenge our strongest assumptions with equally strong arguments, so I'll add my 2c...

Sacrificing extensibility for the sake of ease of adoption isn't entirely desirable, however, giving the user clearly defined examples, following veteran best practices is both valuable to adoption, and useful even to the veteran. We already have two layers: `mkOCI`, and `mkStandardOCI` on top. Perhaps one more general layer (`mkStandardAlpineOCI`?), and maybe even _n_ additional layers for specialized use-cases on top of that would be useful to have, both for use and for reference.

To challenge that a bit though, the biggest failure of DevOS originally was the attempt to encourage this more specialized contribution, and it isn't trivial to _convince_ the domain expert that it is worthwhile to invest their time. In short, it could become a time sync if we take it on ourselves, but with some (future) community adoption it might be trivial.

Still, even seasoned Nix veterans seem to struggle with Standard's model at first, so a wealth of examples may not be such a bad thing. To expand on that, perhaps even the Standard registry could exist in its own dedicated repository. Then maybe it wouldn't be so offensive to add references to officially supported Auxilary functions to plug into Standard, and that don't exist in divnix/std, but elsewhere.

This might even become useful as an actual layer boundary of the on-disk image if we force each function to produce exactly one layer... But then, that's a lot of theory already so I'll leave it there :sweat_smile:

blaggacao:
I think we should aim at MVP-QoL for:
- [ ] existing directories: /tmp, /srv, etc
- [ ] basic /etc: /etc/group, /etc/passwd, etc (+ certs)

---

I think Josh's snipped is a good start.

blaggacao:
further developed in https://github.com/divnix/std/pull/297

PR #275, opened 2023-03-10
refactor: simplify proviso

blaggacao:


nrdxp (changes requested):
I am against this changeset. The proviso should be something that is really quick and doesn't add any noticable time to discovery. This could add tens of seconds or even minutes if we are uploading a lot of images. The purpose of checking them all in parallel is to pay a constant time (the time of the longest round trip) instead of a linear one (k*n)

src/blocktypes/containers.nix
```diff
@@ -69,11 +69,33 @@
           ${copyFn}
           copy docker://${img}
         '';
+        meta.images = map (tag: "${img}:${tag}") tags;
         proviso = l.toFile "container-proviso" ''
-          SKOPEO_INSPECT=${./proviso/skopeo-inspect.sh}
-          ${builtins.readFile ./proviso/publish.sh}
+          function scopeo_inspect() {
+            local images=("$@")
+            for image in ''${images[@]}; do
```
nrdxp, line 76: this is my main issue with this PR. It will dramatically increase the time in takes to check for images, somewhat defeating the purpose of checking to begin with.

blaggacao: Thanks for the review!  😊

Previously, it had been `xargs -P 0`, which I had understood as making it sequential, too. So I had no understanding that this would have effected a regress on performance.

Let me double check, again, where I missed something.

nrdxp: Ah I see, `-P 0` no thread limit, run them all together. combined with `-n 1` which means pass only 1 arg for each invocation, it calls as many instances of skopeo as there are images at once

blaggacao: Reinstated in https://github.com/divnix/std/pull/275/commits/296e7c8c1e10857bcf46fba4bbe2e69c412200fc

blaggacao:
In a sync meeting with @nrdxp we concluded that it is good to merge this and release a `0.20.0` release together with a maintenance branch. But I don't want to interrump his weekend to clear his review. :smile:

Issue #272, opened 2023-02-28
flake-parts module no longer working 


montchr:
I'm using `std` through the flake-parts module, which was working well until a few weeks ago when I started getting the following error after updating inputs:

```console
$ nix develop
error: attribute 'BlockTypes' missing

       at /nix/store/02cdqdy8973dvz5g7ny2sli8nyan8cgm-source/src/flakeModule.nix:70:14:

           69|     descriptionClass = "noun";
           70|     inherit (types.BlockTypes) check;
             |              ^
           71|   };
(use '--show-trace' to show detailed location information)
```

Here is my `flake.nix`:

https://github.com/montchr/dotfield/blob/d9fd62dbf2f6c4f3f8e61cf0b0f87123de720459/flake.nix

I've been busy so have been avoiding the update. In another simpler project I got the same issue but wasn't really using the features of `flake-parts` so I just switched it to use Standard's `mkFlake`.

<details>
<summary>Full Output</summary>

```console
$ nix develop --show-trace
error:
       … while evaluating call site

       at «none»:0: (source not available)

       … while calling anonymous lambda

       at /nix/store/dxfmn2pdvvpiark08jlwc8s8gkhviiw1-source/lib/attrsets.nix:716:24:

          715|     let f = attrPath:
          716|       zipAttrsWith (n: values:
             |                        ^
          717|         let here = attrPath ++ [n]; in

       … while evaluating call site

       at «none»:0: (source not available)

       … while calling 'g'

       at /nix/store/dxfmn2pdvvpiark08jlwc8s8gkhviiw1-source/lib/attrsets.nix:535:19:

          534|           g =
          535|             name: value:
             |                   ^
          536|             if isAttrs value && cond value

       … while evaluating call site

       at /nix/store/dxfmn2pdvvpiark08jlwc8s8gkhviiw1-source/lib/attrsets.nix:538:20:

          537|               then recurse (path ++ [name]) value
          538|               else f (path ++ [name]) value;
             |                    ^
          539|         in mapAttrs g;

       … while calling anonymous lambda

       at /nix/store/dxfmn2pdvvpiark08jlwc8s8gkhviiw1-source/lib/modules.nix:270:72:

          269|           # For definitions that have an associated option
          270|           declaredConfig = mapAttrsRecursiveCond (v: ! isOption v) (_: v: v.value) options;
             |                                                                        ^
          271|

       … while evaluating the attribute 'value'

[...]

PR #276, opened 2023-03-11
fix: prevent flake-parts module scope collision of `types` variables

montchr:
This is another attempt at fixing #272.

~Unfortunately, it looks like there might be another issue uncovered after fixing the initial issue. When using the flake module, I now get this error:~

Also fixed:

```console
error: value is a function while a set was expected

       at /nix/store/l63z3fk6bx37iprrxwgwbzxn55xh0dx3-source/src/flakeModule.nix:69:36:

           68|
           69|     cellBlocksType = mkOptionType {
             |                                    ^
           70|       name = "cellBlocks";
(use '--show-trace' to show detailed location information)
```

### DIFF
--- a/src/flakeModule.nix
+++ b/src/flakeModule.nix
@@ -29,7 +29,6 @@
     mkOptionType
     showDefs
     showOption
-    types
     literalExpression
     ;
 
@@ -56,9 +55,9 @@
     merge = mergeHarvesterOption;
     emptyValue = {value = [];};
     inherit
-      (types.either
-        (types.listOf types.nonEmptyStr)
-        (types.listOf (types.listOf types.nonEmptyStr)))
+      (with lib.types; (either
+        (listOf nonEmptyStr)
+        (listOf (listOf nonEmptyStr))))
       check
       ;
   };
@@ -67,7 +66,11 @@
     name = "cellBlocks";
     description = "list of cell block";
     descriptionClass = "noun";
-    inherit (types.BlockTypes) check;
+    inherit
+      (types.BlockTypes
+        "Block Types usable in the Standard flake-parts module")
+      check
+      ;
   };
 
   opt = options.std;
@@ -83,11 +86,11 @@ in {
            - Find a good [walk-through here](https://jmgilman.github.io/std-book/).
            - And the general [documentation here](https://std.divnix.com/index.html).
         '';
-        type = types.submodule {
+        type = lib.types.submodule {
           options = {
             cellsFrom = mkOption {
               description = "Where Standard discovers Cells from.";
-              type = types.path;
+              type = with lib.types; path;
               example = literalExpression "./nix";
             };
             cellBlocks = mkOption {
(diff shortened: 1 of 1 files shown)


montchr:
Also noting that the flake-parts module do not appear to have test coverage, considering that the tests are still passing.

montchr:
> error: value is a function while a set was expected

Coming from:

https://github.com/divnix/std/blob/a0be894005f5f3987ede606febdf4a698060ed9c/flake.nix#L92-L100

montchr:
Judging by the function's `log` parameter and this usage, I will look into adding a message for logging.

https://github.com/paisano-nix/core/blob/5f2fc05e98e001cb1cf9535ded09e05d90cec131/grow/newExtractFor.nix#LL49C16-L49C21

```nix
      action = types.ActionCommand "Action \"${action'.name}\" of Cell Block \"${cellBlock.name}\" (Cell Block Type: \"${cellBlock.type}\")" action'.command;
```

montchr:
Okay, I think https://github.com/divnix/std/pull/276/commits/f9d73d53a8221f1b79a23853727a78f8774d707b did the trick.

(fwiw I'm still getting errors in my flake but it's been almost a month since I've updated anything due to #272, so I think that's on me)

src/flakeModule.nix
```diff
@@ -4,178 +4,182 @@
...
-      list
-    else throw "Cannot merge definitions of `${showOption loc}', as at least one merge candidate is not a list of strings or a list of list of strings. Definition values:${showDefs defs}";
-
-  harvesterType = mkOptionType {
-    name = "harvester";
-    description = "harvest path(s) as a list of strings or a list of lists of strings";
-    descriptionClass = "noun";
-    merge = mergeHarvesterOption;
-    emptyValue = {value = [];};
+  blockTypes = types.BlockTypes "Block Types usable in the Standard flake-parts module";
```
montchr, line 8: Reformatting (via `alejandra`) caused by the introduction of this variable at the top-level scope, which seemed the most appropriate place to avoid collisions in the lower function scope. But open to other suggestions. 

Initially I took the approach of introducing tightly-scoped usages of `with lib.types;` and removing the shadowed variable. I reverted that once I ran into the second issue fixed in f9d73d53a8221f1b79a23853727a78f8774d707b, but I'm thinking it might still be cleaner to go back to that approach.

montchr, line 8: Also, open to any suggestions for more-appropriate/helpful messaging here.

blaggacao: Upstream I went to great lenghts [here](https://github.com/paisano-nix/core/blob/f71a2db9414d66663c03a65ade97a9f353fb6d55/grow/default.nix#L70-L77) and [here](https://github.com/divnix/std/blob/112cb5d025e344f58e3a4446c4bf034e318ea1bb/flake.nix#L64-L71) (taking great care to not sabotage `unsaceGetAttrsPos`), in order to produce a good `origin` value in `(BlockTypes "${origin "cellBlocks"} - grow[On]:cellBlocks" cellBlocks))`

blaggacao:
This is definetly going to fix the two identified issues. If not a complete fix, we can still follow up. Merging now!

Thank you! :handshake: :partying_face:

bnavetta:
This fixes the issue for me too, thanks @montchr!

PR #278, opened 2023-03-13
fix: don't allow defining broken image

nrdxp:
`recursiveUpdate` takes precedence on the right hand side, and does not merge lists, so the user can overwriting important options set by mk*OCI functions, and end up with a borked image. Instead make the options set by these functions take precedence, and allow `copyToRoot` to be merged cleanly from the one set in `options`.


blaggacao:
Thanks!!

PR #281, opened 2023-03-21
feat(ops): add `revise` family of packages

nrdxp:
Utility functions to allow for building containers in a mono-repo style environment where the source code is contained in the same repository as the std code, specifically so that one may detect meaningful changes to the image via its tag in the special case where the package's output includes the revision of the source code (e.g. for displaying the version to the user).

  Without special processing, this kind of package would cause the OCI image tag to change on each new revision whether the actual contents of the image changed or not. Combined with `std.incl`, one may have a very strong indicator for when the contents of the image actually includes meaningful changes which avoids flooding the remote registry with superflous copies.

  `reviseOCI` can also be called where the package does not need the revsion at build time but you simply want to tag the image by its hash for later processing by the proviso, and you also want to include additional tags on the image, such as the revision.


blaggacao:
@gytis-ivaskevicius @michalrus 

This is PR and the technique behind it is quite relevant for pur repositories where developers expect to see the revision.

This implementation is for OCI, but a similar approach / strategy might be employed for packages / desktop installers and the like.

cells/lib/ops/revise.nix
```diff
@@ -0,0 +1,65 @@
...
+  l = nixpkgs.lib // builtins;
+in
+  /*
+  For use with `revisePackage` and `reviseOCI` to build containers in a mono-repo style
+  environment where the source code is contained in the same repository as the std code,
+  specifically so that one may detect meaningful changes to the image via its tag in the
+  special case where the package's output includes the revision of the source code (e.g. for
+  displaying the version to the user).
+
+  Without special processing, this kind of package would cause the OCI image tag to change
```
blaggacao, line 15:
```
Explanation at the example of OCI Images:
```

To make it clear this specificity is at the service of an example, not a tacit narrowing of the api (as `revisePackage` was mentioned explicitly before).

cells/lib/ops/revise.nix
```diff
@@ -0,0 +1,65 @@
...
+  copies.
+
+  Args:
+  op: Optional function which takes the package as an argument.
+  pkg: The package you wish to revise.
+  fn: Optional functor with a reference to `pkg` if needed by `op`.
+
+  Returns:
+  The package with a clone of itself in the passthru where the expected revision is set to
+  "not-a-commit" for later use by `reviseOCI`.
```
blaggacao, line 28:
```
and `revisePackage`
```

cells/lib/ops/revise.nix
```diff
@@ -0,0 +1,65 @@
...
+  */
+  op: pkg: fn: let
+    result = op (fn pkg);
+    dummy = "not-a-commit";
+    rev = pkg.src.rev or pkg.src.origSrc.rev or dummy;
+  in
+    if pkg ? sansrev || (rev != dummy && result == pkg)
```
blaggacao, line 35: `sansrev` looks like a contract that could be documented. If not public, I, personally would benefit from a couple of code comments below.

cells/lib/ops/revise.nix
```diff
@@ -0,0 +1,65 @@
+{
+  inputs,
+  cell,
+}: let
+  inherit (inputs) nixpkgs;
+  l = nixpkgs.lib // builtins;
+in
+  /*
+  For use with `revisePackage` and `reviseOCI` to build containers in a mono-repo style
+  environment where the source code is contained in the same repository as the std code,
```
blaggacao, line 10: I think that is an unnecessary restriction.

This helps whenever unrelated changes should not bump the `rev` of a build.

cells/lib/ops/reviseOCI.nix
```diff
@@ -0,0 +1,63 @@
+{
+  inputs,
+  cell,
+}: let
+  inherit (inputs) nixpkgs;
+  l = nixpkgs.lib // builtins;
+in
+  /*
+  Utility function to allow for building containers in a mono-repo style environment where
+  the source code is contained in the same repository as the std code, specifically so that
```
blaggacao, line 10: As above: unecessary restriction?

cells/lib/ops/revise.nix
blaggacao: Or generalize the explanation since `reviseOCI` has the same parragraph...

cells/lib/ops/reviseOCI.nix
```diff
@@ -0,0 +1,63 @@
...
+  with `std.incl`, one may have a very strong indicator for when the contents of the image
+  actually includes meaningful changes which avoids flooding the remote registry with superflous
+  copies.
+
+  This function can also be called where the package does not need the revsion at build time
+  but you simply want to tag the image by its hash for later processing by the proviso, and
+  you also want to include additional tags on the image, such as the revision.
+
+  Args:
+  args: arguments to the mkOCI function to be called.
```
blaggacao, line 26: Args is the whole caller args, now (`args @`). Maybe that's been refactored? Imo the `...` covers it well.

cells/lib/ops/reviseOCI.nix
```diff
@@ -0,0 +1,63 @@
...
+  args @ {
+    operable,
+    mkOCI ? cell.ops.mkStandardOCI,
+    ...
+  }: let
+    revision = cell.ops.revise mkOCI args.operable (operable: builtins.removeAttrs (args // {inherit operable;}) ["mkOCI"]);
+  in
+    if args.operable ? sansrev
+    then
+      mkOCI (args
```
blaggacao, line 45: Probably an `args'` was intended with `mkOCI` stripped.

cells/lib/ops/reviseOCI.nix
```diff
@@ -0,0 +1,63 @@
...
+            };
+        })
+    else
+      mkOCI (args
+        // {
+          meta =
+            args.meta
+            or {}
+            // {
+              tags = [(cell.ops.hashOfPath revision.outPath)] ++ (args.meta.tags or []);
```
blaggacao, line 61: Hm, now I wonder even more how the magic of `sansrev` signalling and dry-running works.

Maybe `revise.nix` would indeed be a good place to explain.

cells/lib/ops/revisePackage.nix
```diff
@@ -0,0 +1,34 @@
...
+  environment where the source code is contained in the same repository as the std code,
+  specifically so that one may detect meaningful changes to the image via its tag in the
+  special case where the package's output includes the revision of the source code (e.g. for
+  displaying the version to the user).
+
+  Without special processing, this kind of package would cause the OCI image tag to change
+  on each new revision whether the actual contents of the image changed or not. Combined
+  with `std.incl`, one may have a very strong indicator for when the contents of the image
+  actually includes meaningful changes which avoids flooding the remote registry with superflous
+  copies.
```
blaggacao, line 19: We could explain how this applies to packages here ☺️

cells/lib/ops/revisePackage.nix
```diff
@@ -0,0 +1,34 @@
...
+
+  Args:
+  target: Same as the first argument to upstream `callPackage`.
+  args: Arguments to `callPackage`.
+
+  Returns:
+  The package with a clone of itself in the passthru where the expected revision is set to
+  "not-a-commit" for later use by `revise` & `reviseOCI`.
+  */
+  target: args @ {rev, callPackage ? nixpkgs.callPackage, ...}: let
```
blaggacao, line 29: Is there a use-case in making `callPackages` overridable? I know the `callPackage` argument API is cheap via global context (which itself if expensive). 

But still: reading the code here, I can't but wonder about why having it overridable.

cells/lib/ops.nix
```diff
@@ -3,8 +3,10 @@
   cell,
 }: let
   inherit (inputs.cells.std.errors) requireInput;
-  inherit (import "${inputs.self}/deprecation.nix" inputs) warnWriteShellEntrypoint;
```
blaggacao, line 6: Thx! 😊

blaggacao (commented):
A first pass, focusing on understanding and trying to ingest the docstrings.

I still have some knowledge gaps on the detailed working of `sansrev`.

I understand the dry-run capability and how it's strategically leveraged. But I don't understand the implementation (yet) 😢.

That's what the next round will be for, hopefully 🤞

cells/lib/ops/revise.nix
nrdxp: well the order as far as the user is concerned is `revisePackage` -> `revise` (on the operable) -> `reviseOCI`, so its speaking to that sequence. The use of revise in `revisePackage` is strictly an implementation detail

cells/lib/ops/revise.nix
nrdxp: the section after the `||` is to tighten the contract from `revisePackage`. The first contract (existence of sansrev) is universal but the second case is only valid in ussage of revisePackage and so that is made explicit since this condition can only be met if the two functors passed in are both the `id` function.

cells/lib/ops/revise.nix
nrdxp: No it really doesn't, this information is only useful for the image tag. The Hash is going to change if the rev changes if the rev is included in the output. The only way to avoid that is to not pass in the revision at all.

cells/lib/ops/reviseOCI.nix
nrdxp: Unless you can give me an example I'm not thinking of, I don't think so

cells/lib/ops/reviseOCI.nix
nrdxp: yep, fixed in a force push

cells/lib/ops/reviseOCI.nix
nrdxp: It has two uses, I think both spelled out here. What other usecases could you imagine?

cells/lib/ops/reviseOCI.nix
nrdxp: The whole point of sansrev is to have a version of the package that doesn't change when the revision does. The only way to guarantee this is to take the rev as an argument to callPackage (hence why `rev` is a required argument to `revisePackage`), and pass in the default `not-a-commit` unconditionally to it.

I tried more clever approaches like modifying `src.origSrc.rev` but the hash of the outPath hash of the src is different so that didn't work.

cells/lib/ops/revisePackage.nix
nrdxp: I've struggled to explain this from the beginning, so if you have any suggestions I'm all for it :sweat_smile:

cells/lib/ops/revise.nix
blaggacao: Does this mean `revidePackage` can't be used standalone for this purpose?

cells/lib/ops/revise.nix
blaggacao: One use case may be a published installer of something in a monorepo that also needs its most significant revision baked into it.

blaggacao:
```
Allows to find the most significant commit in a monorepo where the source is filtered, but still needs a revision. For use with `revisePackage` and `reviseOCI`.
```

cells/lib/ops/reviseOCI.nix
blaggacao: Thanks, the overall strategy is clear to me _now_, but I'm more advocating for code comments on behalf of a future self or present/future other.

Can you curry and spice up the code with a couple of code comments that guide the _code_ reader through the logic of `sansrev`?

cells/lib/ops/revisePackage.nix
blaggacao: I'm afraid that `callPackage` is not a public api on targets, at all. Targets of type `package` are already of type `derivation`.

blaggacao: I think what you mean is `pkgFun`, that aren't yet of block type `packages`.

cells/lib/ops/revise.nix
blaggacao: I've made an idempotent refactoring  and I don't understand what this function does:
```nix
  op: pkg: fn: let
    result = op (fn pkg);
    dummy = "not-a-commit";
    rev = pkg.src.rev or pkg.src.origSrc.rev or dummy;

    sansrev = let
      pkg' = op (fn (pkg.sansrev or (pkg.override {rev = dummy;})));
      passthru = pkg'.passthru or {} // {outHash = cell.ops.hashOfPath pkg'.outPath;};
    in
      pkg'.overrideAttrs (_: {inherit passthru;});
    passthru = result.passthru or {} // {inherit sansrev;};
  in
    if pkg ? sansrev || (rev != dummy && result == pkg)
    then result.overrideAttrs (_: {inherit passthru;})
    else result
```

Does it do more than one thing, and if so how many?

blaggacao: I gain the impression this function could also be called `42` :smile:

nrdxp: It's called like `revise mkOerable packages.operable (operable: { /* ... */ })` The functor is necessary to defer the binding of `operable`

PR #283, opened 2023-03-23
fix build proviso

blaggacao:
- fix: use nix-uncached with optimizations and fixes
- fix: query correct targets for uncached items


GTrunSec:
Could you update-lock for the inputs.paisano?

blaggacao:
> Could you update-lock for the inputs.paisano as well?

@GTrunSec done in #284

PR #286, opened 2023-03-24
fix: build proviso needs to find uncached paths

nrdxp:
"will be fetched" catches packages that are already cached, we need to
find those that aren't.


src/build-fetched-extractor.sed
```diff
@@ -1,5 +1,5 @@
 # If the line 'will be fetched' is matched ...
```
blaggacao, line 1:
```suggestion
# If the line 'will be built' is matched ...
```

PR #288, opened 2023-03-27
(added (readme (proposal legacyCellApi)))

LiGoldragon:
This type of API would make `std` useful enough for kriomOS to adopt it.
You can watch [this youtube live](https://youtube.com/live/e8AQy8EBAzE?feature=share) to get a closer view of my opinion and solutions.

Thanks, it is beautiful code. Well done.
Gratefully
Li Goldragon

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,9 @@ But hey! It means: we can progress together!
     #    Run for example: 'mkdir nix/mycell'
     # 2. Each <block>.nix or <block>/default.nix within it becomes a "Cell Block"
     #    Run for example: '$EDITOR nix/mycell/packages.nix' - see example content below
-    cellsFrom = ./nix;
+    cellsFrom = ./nix/cells;
+    overlaysFrom = ./nix/overlays/;
+    packagesFrom = ./nix/packages/;
     # 3. Only blocks with these names [here: "packages" & "devshells"] are picked up by Standard
     #    It's a bit like the output type system of your flake project (hint: CLI & TUI!!)
     cellBlocks = with std.blockTypes; [


LiGoldragon:
Could this type of PR get accepted into `std`?
Thanks!

blaggacao:
Hi @LiGoldragon !

Sorry for the late reply, I've been focusing on a [bunch](https://github.com/paisano-nix/direnv) of [other things](https://github.com/paisano-nix/onboarding).

I'm on the fence of expanding the top level api. But I have reasons and solutions for you:

Overlays mutatate something that is fully qualified and well known (`inputs.nixpkgs`). From one project to another that creates non-portable private knowledge that also is surprising (unless overlays are one's second nature).

To address this, but still provide the power of overlays, the `blockTypes.pkgs` was introduced. It doesn't register into the registry and, thus, doesn't show up in CI or the CLI (which wouldn't be useful nor intended for the entire nix package collection).

Thus, the pattern becomes simply: `./nix/toolchain/pkgs.nix`, with content:
```nix
inputs.nixpkgs.appendOverlays []
```

This is a full equivalent using the same code path as if reimporting (and also has the same eval cost).

As for `packagesFrom`, I've been just yesterday playing with [Haumea](https://github.com/nix-community/haumea). I beleive, this is the way to go. A pattern could look like:
`nix/toolchain/mypkgs{/,.nix}` where the file would have as content something like (untested):

```nix
inputs.haumea.lib.load {
  src = ./mypkgs;
}
```

I hope that these are good (or even better) solutions. 😊

Thanks for taking the effort to analyze Standard in depth and to improve it! 🤝

LiGoldragon:
This is an amazing response. No need to be sorry. I came across haumea right after my last comment, through your `hive` repo. I will be putting my work related to this in [flakeWorld](https://github.com/sajban/flakeWorld).

  I will develop an interface that exposes both `std` along with a legacy-based one (pkgs). You're right about overlays, and I was avoiding them for years until I was pulled back into it recently, but it's only causing me headaches, so I'm going to reverse it.
  
   You can watch my live coding but I think it's not your style, it'll be too slow for you. Im very curious to know you, your code is beautiful. Would love to do an interview, if you're into this. 
   
   Drink the Water of Shiva, and be forever free from disease and starvation.
   May your feelings be as light and bright as a flame, and as gentle and embracing as water.

PR #296, opened 2023-05-09
docs: fix emojis in mdbook

figsoda:
the `:heart_eyes:` emojis are specific to github and shows up as plain text on the website
image


blaggacao:
Thank you so much for the attention to detail that sometimes exceeds what I've been capable of (in this case after a long long day :smile:)

PR #297, opened 2023-05-13
feat: improve std image

blaggacao:
adding /tmp and certs for scoring in a higher bucket of the frequenctly-required-by-users histogram (armchair estimate / educated guess)

set up linux pam files for the nobody / nogroup user

### DIFF
--- a/cells/lib/ops/mkStandardOCI.nix
+++ b/cells/lib/ops/mkStandardOCI.nix
@@ -2,7 +2,7 @@
   inputs,
   cell,
 }: let
-  inherit (inputs) nixpkgs;
+  inherit (inputs) nixpkgs dmerge;
   l = nixpkgs.lib // builtins;
   n2c = inputs.n2c.packages.nix2container;
 in
@@ -40,16 +40,22 @@ in
     options ? {},
     meta ? {},
   }: let
+    inherit (operable) passthru;
+    inherit (operable.passthru) livenessProbe readinessProbe runtimeInputs runtime debug;
+
+    hasLivenessProbe = passthru ? livenessProbe;
+    hasReadinessProbe = passthru ? readinessProbe;
+    hasDebug = args.debug or false;
+
     # Link useful paths into the container.
-    runtimeEntryLink = let
-      runtimeShell = l.getExe operable.passthru.runtime;
-    in ''
+    runtimeShell = l.getExe runtime;
+    runtimeEntryLink = ''
       ln -s ${runtimeShell} $out/bin/sh
       ln -s ${runtimeShell} $out/bin/runtime
     '';
-    debugEntryLink = l.optionalString debug "ln -s ${l.getExe operable.passthru.debug} $out/bin/debug";
-    livenessLink = l.optionalString (operable.passthru ? livenessProbe) "ln -s ${l.getExe operable.passthru.livenessProbe} $out/bin/live";
-    readinessLink = l.optionalString (operable.passthru ? readinessProbe) "ln -s ${l.getExe operable.passthru.readinessProbe} $out/bin/ready";
+    debugEntryLink = l.optionalString hasDebug "ln -s ${l.getExe debug} $out/bin/debug";
+    livenessLink = l.optionalString hasLivenessProbe "ln -s ${l.getExe livenessProbe} $out/bin/live";
+    readinessLink = l.optionalString hasReadinessProbe "ln -s ${l.getExe readinessProbe} $out/bin/ready";
 
     # Wrap the operable with sleep if debug is enabled
     debugOperable = cell.ops.writeScript {
@@ -64,7 +70,7 @@ in
       '';
     };
     operable' =
-      if debug
+      if hasDebug
       then debugOperable
       else operable;
 
@@ -75,30 +81,32 @@ in
       ${livenessLink}
       ${readinessLink}
     '';
-  in
-    cell.ops.mkOCI (
-      {
-        inherit name uid gid labels options perms config meta;
-        entrypoint = operable';
-        setup = [setupLinks] ++ setup;
-        runtimeInputs = operable.passthru.runtimeInputs;
 
-        # Put liveness and readiness probes in a separate layer
-        layers = [
-          (n2c.buildLayer {
-            deps =
-              []
-              ++ (l.optionals (operable.passthru ? livenessProbe) [(n2c.buildLayer {deps = [operable.passthru.livenessProbe];})])
-              ++ (l.optionals (operable.passthru ? readinessProbe) [(n2c.buildLayer {deps = [operable.passthru.readinessProbe];})]);
-            maxLayers = 10;
-          })
-        ];
-      }
-      // l.throwIf (args ? tag && meta ? tags)
+    tag' =
+      l.throwIf (args ? tag && meta ? tags)
       "mkStandardOCI: use of `tag` and `meta.tags` arguments are not supported together. Remove the former."
-      (
-        l.optionalAttrs (tag != "") ((import "${inputs.self}/deprecation.nix" inputs).warnLegacyTag {
-          inherit tag;
-        })
+      l.optionalString (tag != "") ((import "${inputs.self}/deprecation.nix" inputs).warnLegacyTag tag);
+  in
+    with dmerge;
+      cell.ops.mkOCI (
+        merge
+        {
+          inherit name uid gid labels options perms config meta setup runtimeInputs;
+          entrypoint = operable';
+          tag = tag'; # ideally ''
+        }
+        {
+          # mkStandardOCI differentiators over mkOCI
+          # - live & readiness probes
+          layers = [
+            # Put liveness and readiness probes in a separate layer
+            (n2c.buildLayer {
+              maxLayers = 10;
+              deps =
+                (l.optionals hasLivenessProbe [(n2c.buildLayer {deps = [livenessProbe];})])
+                ++ (l.optionals hasReadinessProbe [(n2c.buildLayer {deps = [readinessProbe];})]);
+            })
+          ];
+          setup = prepend [setupLinks];
+        }
       )
-    )
--- a/cells/lib/ops/mkStandardOCI.nix
+++ b/cells/lib/ops/mkStandardOCI.nix
@@ -74,12 +74,36 @@ in
       then debugOperable
       else operable;
 
-    setupLinks = cell.ops.mkSetup "links" [] ''
-      mkdir -p $out/bin
-      ${runtimeEntryLink}
-      ${debugEntryLink}
-      ${livenessLink}
-      ${readinessLink}
+    inherit (nixpkgs.dockerTools) caCertificates;
+    setupLinks =
+      cell.ops.mkSetup "links" [
+        {
+          regex = "/bin";
+          mode = "0555";
+        }
+      ] ''
+        mkdir -p $out/bin
+        ${runtimeEntryLink}
+        ${debugEntryLink}
+        ${livenessLink}
+        ${readinessLink}
+      '';
+
+    users = cell.ops.mkUser {
+      user = "nobody";
+      group = "nogroup";
+      uid = "65534";
+      gid = "65534";
+      withRoot = true;
+      withHome = true;
+    };
+
+    tmp = nixpkgs.runCommand "tmp" {} ''
+      mkdir -p $out/tmp
+    '';
+
+    nss = nixpkgs.writeTextDir "etc/nsswitch.conf" ''
+      hosts: files dns
     '';
 
     tag' =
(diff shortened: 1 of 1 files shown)


blaggacao:
@nrdxp @jmgilman @gytis-ivaskevicius 

I'd like to have a thorough discussion of this as to not to screw up the API here. So let's not take this lightly. Couple of initial questions below.

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,19 +68,43 @@ in
...
-      ${runtimeEntryLink}
-      ${debugEntryLink}
-      ${livenessLink}
-      ${readinessLink}
-    '';
+    setupLinks =
+      cell.ops.mkSetup "links" [
+        {
+          regex = "/bin";
+          mode = "0555";
```
blaggacao, line 75: here this means `rx-rx-rx` on `/bin`.

A suitable restriction?

Is there an escape hatch? Yes: `[{regex = "/bin/mything"; mode = "0100";}] ":"`

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,19 +68,43 @@ in
...
+    setupBareMinimumEnv =
+      cell.ops.mkSetup "env" [
+        {
+          regex = "/tmp";
+          mode = "0777";
+        }
+      ] ''
+        mkdir -p $out/tmp
+        # Setup certificates
+        cp -r ${nixpkgs.cacert}/* $out
```
blaggacao, line 94: `mkStandardOCI` serving P95, do we have indication or suspicion that this would break within the 95th percentile of usage instances/scenarios? (since we're in lack of data, this is armchair guesswork)

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,19 +68,43 @@ in
...
+      '';
+
+    setupSparseUser = cell.ops.mkUser {
+      user = "nobody";
+      group = "nogroup";
+      inherit uid gid;
+    };
```
blaggacao, line 101: Similarly is there a penality within P95 for not exposing `withHome` & `withRoot`. Please expend the effort and[ look up the details](https://github.com/divnix/std/blob/main/cells/lib/ops/mkUser.nix#L30-L71) of what they do.

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,19 +68,43 @@ in
...
   in
     cell.ops.mkOCI (
       {
         inherit name uid gid labels options perms config meta;
         entrypoint = operable';
-        setup = [setupLinks] ++ setup;
+        setup = [setupLinks setupBareMinimumEnv] ++ setup;
```
blaggacao, line 107: I refrained from adding `/etc`, by design. However, it's symlinked by coincidence via `nixpkgs.cacert`.

But erring on the side of design, I consider it may be slightly below P95 already.

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,20 +68,82 @@ in
...
+        # docker creates some minimal files in /etc at runtime
+        # that would otherwise shadow a top-level symlink:
+        #   bash-5.2$ ls /etc
+        #   hostname  hosts  mtab  resolv.conf
+        # therefore, we make trick nixpks.buildenv into not
+        # generating symlinks at the level of /etc but at least
+        # one level deeper by artifie of this overlapping element
+        # that buildenv will detect and push the linking one level down
+      REASON
+    '';
```
blaggacao, line 102: @gytis-ivaskevicius

blaggacao: but also @nlewo

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,20 +68,82 @@ in
...
+        config =
+          config
+          // {
+            Env =
+              config.Env
+              or []
+              ++ [
+                # Adopted from: <nixpks>/pkgs/data/misc/cacert/setup-hook.sh
+                # This file is created when nixpkgs.cacert is linked to the root
+                "NIX_SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt"
```
nrdxp, line 130: Is this only for the `nix` bin? We probably don't want that in the container most times.

blaggacao: True! Its existence might confuse unnecessarily.

cells/lib/ops/mkStandardOCI.nix
nrdxp: There is a `dockerTools.caCertificates`, but I'm not sure if there is any difference.

gytis-ivaskevicius: yeah, I tried `dockerTools.caCertificates`. Result seemed identical

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -68,20 +68,82 @@ in
...
+      mkdir -p $out/tmp
     '';
   in
     cell.ops.mkOCI (
       {
-        inherit name uid gid labels options perms config meta;
+        inherit name uid gid labels meta;
         entrypoint = operable';
-        setup = [setupLinks] ++ setup;
+        setup = [setupLinks setupCerts setupEtcHack] ++ setup;
```
gytis-ivaskevicius, line 118: Isn't this forcing all images to have certs?

cells/lib/ops/mkStandardOCI.nix
gytis-ivaskevicius: I am not sure how this changes anything. Is it a special exception in `buildEnv` function?

cells/lib/ops/mkStandardOCI.nix
blaggacao: Indeed, one reasoning may be that `mkStandardOCI` comes with the `curl` certificate authorities.

That's not what high assurance environments (e.g. medicine, military, transport, etc) would want, but there's `mkOCI`, still.

cells/lib/ops/mkStandardOCI.nix
blaggacao: `buildEnv` has a shallow-first linking strategy, when we inhibit top level linking by this fake entropy, `buildEnv` will link one level down, instead.

cells/lib/ops/mkStandardOCI.nix
blaggacao: @gytis-ivaskevicius Can you maybe confirm if [this nixpkgs example](https://github.com/NixOS/nixpkgs/blob/d925734d3bb7f12924e6016cd33222684b5435f5/pkgs/build-support/docker/examples.nix#L727) suffers from the same issue on local docker?

gytis-ivaskevicius: @blaggacao 
```
bash-5.2# ls -la /
total 0
drwxrwxr-x   1 0 0   88 May 16 09:15 .
drwxrwxr-x   1 0 0   88 May 16 09:15 ..
-rwxr-xr-x   1 0 0    0 May 16 09:15 .dockerenv
dr-xr-xr-x   1 0 0 1040 Jan  1  1980 bin
drwxr-xr-x   5 0 0  360 May 16 09:15 dev
dr-xr-xr-x   1 0 0   68 May 16 09:15 etc
dr-xr-xr-x   1 0 0    8 Jan  1  1980 lib
dr-xr-xr-x   1 0 0   18 Jan  1  1980 libexec
dr-xr-xr-x   1 0 0   10 Jan  1  1980 nix
dr-xr-xr-x 600 0 0    0 May 16 09:15 proc
dr-xr-xr-x   1 0 0    6 Jan  1  1980 share
dr-xr-xr-x  13 0 0    0 May 15 15:28 sys
bash-5.2# ls -la /etc
total 16
dr-xr-xr-x 1 0 0  68 May 16 09:15 .
drwxrwxr-x 1 0 0  88 May 16 09:15 ..
-rw-r--r-- 1 0 0  13 May 16 09:15 hostname
-rw-r--r-- 1 0 0 174 May 16 09:15 hosts
lrwxrwxrwx 1 0 0  12 May 16 09:15 mtab -> /proc/mounts
dr-xr-xr-x 1 0 0   6 Jan  1  1980 pki
-rw-r--r-- 1 0 0 974 May 16 09:15 resolv.conf
dr-xr-xr-x 1 0 0  10 Jan  1  1980 ssl
bash-5.2# ls -la /etc/ssl/certs
total 8
dr-xr-xr-x 1 0 0 64 Jan  1  1980 .
dr-xr-xr-x 1 0 0 10 Jan  1  1980 ..
lrwxrwxrwx 1 0 0 87 Jan  1  1980 ca-bundle.crt -> /nix/store/7vwvn3d5yb5yb3vdv9iv4321ckx4paf9-nss-cacert-3.86/etc/ssl/certs/ca-bundle.crt
lrwxrwxrwx 1 0 0 87 Jan  1  1980 ca-certificates.crt -> /nix/store/7vwvn3d5yb5yb3vdv9iv4321ckx4paf9-nss-cacert-3.86/etc/ssl/certs/ca-bundle.crt
```

cells/lib/ops/mkStandardOCI.nix
gytis-ivaskevicius: I'd be okay with having this as default since it does not take any extra space/build time
What I am wondering, doesn't NIX_SSL_CERT_FILE default to this location?

cells/lib/ops/mkStandardOCI.nix
gytis-ivaskevicius: I am not sure how I feel about it but I am happy that it resolves my issue

cells/lib/ops/mkStandardOCI.nix
blaggacao: Thanks a lot!

At the same time, this is strange. `coreutils` and `caCertificates` only account for a portion of the observable file system hierarchy.

Where would the rest come from?

cells/lib/ops/mkStandardOCI.nix
blaggacao: It's a cutoff criteria that we need to apply here, and it's indeed not trivial to strike a good balanace.

cells/lib/ops/mkStandardOCI.nix
blaggacao: > What I am wondering, doesn't NIX_SSL_CERT_FILE default to this location?

In the newer version of the branch, I removed it and I'm using `dockerTools.caCertificates` instead.

This package explains that, unlike the `cacert` package directly, it copies the certs to the canonical locations  for various distros.

However, it omits to set the env variable for, e.g. the haskell ssl implememtation, which does not implement a canonical location, but requires the explicit hint from the environment (what a good choice!).

For that reason, I left openssl amd Haskell compat, even if the former would have otherwise a canonical detection depending on the distro's build of it.

This ends up a precedence for setting env variables for tools and language systems which implement SSL, so I left it, but without unnecessarily hinting at Nix being a special kid.

cells/lib/ops/mkStandardOCI.nix
blaggacao: This left me still wondering if `caCertificates` is autonomously solving the top level linking problem or if some external effect of the particular environment is at play.

The bug is only really fixed if `caCertificates` all on its own is capable of dodging the issue with top level linking of `/etc` and docker's, then destructive, overlay files in `/etc`.

gytis-ivaskevicius: > Where would the rest come from?
coreutils contains `bin`, `share` and `libexec`
bash contains `bin` and `lib`
`dockerTools.caCertificates` contains `etc`

cells/lib/ops/mkStandardOCI.nix
```diff
@@ -64,41 +70,85 @@ in
...
+      '';
+
+    users = cell.ops.mkUser {
+      user = "nobody";
+      group = "nogroup";
+      uid = "65534";
+      gid = "65534";
+      withRoot = true;
+      withHome = true;
+    };
```
gytis-ivaskevicius, line 99: It might makes sense to add `root` and `app` (default application runner) users as well

gytis-ivaskevicius (approved):

cells/lib/ops/mkStandardOCI.nix
blaggacao: Root is added by `withRoot`, already.

`nobody` & `nogroup` is the app user signaling that they are not supposed to signify any impersonation, even if they do on the high number `65534`, in practice.

Unless the app has a wired dependency on some additional entries in nss,  `/etc/pass` & `/etc/group`, this is a good baseline.

Should the app have such a "wired" (i.e. presumptive) dependency, it's probably best to make this explicit to the implementor.

PR #305, opened 2023-06-04
fix: add prj-spec dirs to gitignore

montchr:
On a fresh update of my local copy of the repo,
`.data/` was listed in the untracked files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,14 @@
 /**/flake.lock
 result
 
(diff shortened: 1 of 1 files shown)


blaggacao (approved):
Thank you! Prj-Specbis supposed to amend `.git/info/exclude` (iirc), but this is better. I need to ammend the wip impl of prj spec

PR #310, opened 2023-06-08
Fix command

whs-dot-hk:


whs-dot-hk:
Hey, we started using std and std-action and find it very useful, thanks

Issue #318, opened 2023-06-26
Std `ops.mkStandardOCI` compatibility problem with `ops.mkOperable` to wrap


Pegasust:
I have a pretty standard poetry2nix web application that I wrap on its environment derivation using `hypercorn`. The built OCI fails to detect the correct binary location because it relies on `lib.getExe`. `mkOperable` works as expected, though.

# Symptoms

I don't have the exact error that throws from Docker engine. It should be something along the line of Docker container not being to start up because `/bin/entrypoint` not found. 

If you have access to `ls` (I use Python's `os.listdir` because I'm too lazy to build a dev layer), You would see that `/bin/entrypoint` exists, but it is a broken symlink.

```console
$ std //ops/oci/p2n-env-oci:load
docker-daemon:<image>:<hash>
$ docker run -it --rm --entrypoint /bin/sh <image>:<hash>
$ python3
```

```py
import os

def follow_symlinks(path):
    if os.path.islink(path):
        link_path = os.readlink(path)
        print(f'{path} -> {link_path}')
        return follow_symlinks(link_path)
    elif os.path.exists(path):
        print(f'Path is a real file or directory: {path}')
        return path
    else:
        print(f'Path not found: {path}')
        return None

follow_symlinks('/bin/entrypoint')
```

One example output is

```console
/bin/entrypoint -> /nix/store/<hash>-oci-setup-links/bin/entrypoint
/nix/store/<hash>-oci-setup-links/bin/entrypoint -> /nix/store/<hash>-operable-python3-3.10.11-env/bin/operable-python3
Path not found: /nix/store/<hash>-operable-python3-3.10.11-env/bin/operable-python3
>>> os.listdir('/nix/store/<hash>-operable-python3-3.10.11-env/bin/')
["operable-python3-3.10.11-env"]
```

# Workaround

```nix
  p2n-env-fix = workspace_app.dependencyEnv.overrideAttrs (prev: {
    # HACK: This allows mkOperable and mkStandardOCI to resolve symlink correctly
    name = "p2n-env-fix";
    meta.mainProgram = "python3";
  });
```

# Nix flakes and its workaround variant

```nix
# nix/dev/packages/tools.nix
{
  inputs,
  cell,
}: let
  system = inputs.nixpkgs.system;
  poetry2nix = inputs.nix-boost.pkgs.${system}.mypkgs.poetry2nix;
  workspace_root = "${inputs.self}";
  workspace_app = poetry2nix.mkPoetryApplication {
[...]

PR #331, opened 2023-07-13
fix(lib/ops): ensure the getName contract is uphold for operables

blaggacao:
fix: #318

# Context

Operables wrap packages, and then they are symlinked via `lib.getExe` into
the container.

```nix
  /* Get the path to the main program of a derivation with either
     meta.mainProgram or pname or name

     Type: getExe :: derivation -> string

     Example:
       getExe pkgs.hello
       => "/nix/store/g124820p9hlv4lj8qplzxw1c44dxaw1k-hello-2.12/bin/hello"
       getExe pkgs.mustache-go
       => "/nix/store/am9ml4f4ywvivxnkiaqwr0hyxka1xjsf-mustache-go-1.3.0/bin/mustache"
  */
  getExe = x:
    "${lib.getBin x}/bin/${x.meta.mainProgram or (lib.getName x)}";
```

Prior to this commit, using `package.name` (which is the wrong attribute, anyways),
we where furnishing that name to `/bin/<name>`.
While `lib.getBin` reads the store path verbatim and thereby honors that name,
`lib.getName` resolves differently via the following implementation resulting
in naming mismatches for the call to `getExe` which ultimately linked
`/bin/entrypoint` in mkOCI to a non-existing exectuable:

```nix
  /* This function takes an argument that's either a derivation or a
     derivation's "name" attribute and extracts the name part from that
     argument.

     Example:
       getName "youtube-dl-2016.01.01"
       => "youtube-dl"
       getName pkgs.youtube-dl
       => "youtube-dl"
  */
  getName = x:
   let
     parse = drv: (parseDrvName drv).name;
   in if isString x
      then parse x
      else x.pname or (parse x.name);
```

# Solution

Feed to the consumer (`getExe`) it's very own contract (`getName`)

---

@Pegasust would you have the ability to test this fix in your particular situation?

### DIFF
--- a/src/lib/ops/mkOperable.nix
+++ b/src/lib/ops/mkOperable.nix
@@ -76,7 +76,7 @@ in
         cell.ops.writeScript
         ({
             inherit runtimeInputs runtimeEnv;
-            name = "operable-${package.name}";
+            name = "operable-${l.getName package}";
             text = ''
               ${runtimeScript}
             '';
--- a/tests/_snapshots/cells-lib-ops
+++ b/tests/_snapshots/cells-lib-ops
@@ -4,7 +4,7 @@
   mkDevOCI = <derivation image-hello-dev.json>;
   mkMicrovm = <derivation microvm-qemu-nixos>;
   mkOCI = "missing-test";
-  mkOperable = <derivation operable-hello-2.12.1>;
+  mkOperable = <derivation operable-hello>;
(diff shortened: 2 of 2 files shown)


Pegasust:
Will look into this

Pegasust:
Confirmed that this solves my use case for #318

Issue #333, opened 2023-08-09
mkDevOCI doesn't work


GTrunSec:
can be sure the problem caused by https://github.com/divnix/std/commit/a736f37d959c234501c5ddc704a283bf1645114d#diff-6fc5688f9ea3cf1e83cb9cbc47d0fd9f9a0250c13c1b0c1bc584a2a799473cafR73

# reproduce the issue
command: 

```
podman run --platform linux/amd64 -it -v "$(pwd):/work" --userns=keep-id --entrypoint /bin/bash d56510d511f2
```

output:
```
Error: crun: executable file `/bin/bash` not found in $PATH: No such file or directory: OCI runtime attempted to invoke acommand that was not found
```

PR #334, opened 2023-08-09
fix(ops/mkDevOCI): avoid undue symlink and improve ux

GTrunSec:
fixes: #333 

It appears that putting those two stages together would result in the entrypoints being overwritten.

### DIFF
--- a/src/lib/ops/mkOCI.nix
+++ b/src/lib/ops/mkOCI.nix
@@ -81,7 +81,7 @@ in
                 paths =
                   setup
                   ++ [
-                    # prevent the $out`/bin` to be a symlink
+                    # trick `buildEnv` and prevent the $out`/bin` to be a symlink
                     (nixpkgs.runCommand "setupDirs" {}
                       ''
                         mkdir -p $out/bin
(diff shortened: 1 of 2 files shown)


blaggacao:
Is there a way this can be resolved specifically via package priority instead? It is clear that a more _special_ `/bin/entrypoint` from `mkDevOCI` should have a higher priority than the generic `/bin/entrypoint` from `mkOCI`.

Deviating the code-path from `buildEnv` might have the unintended side effect of not properly managing priorities in other cases where that behaviour is intended.

In other words, this fix breaks the API and I'd prefer one that doesn't.

src/lib/ops/mkOCI.nix
```diff
@@ -78,7 +78,16 @@ in
             [
               (nixpkgs.buildEnv {
                 name = "root";
-                paths = [setupLinks] ++ setup;
+                paths =
+                  setup
+                  ++ [
+                    # prevent the $out`/bin` to be a symlink
```
blaggacao, line 84:
```suggestion
                    # trick `buildEnv` and prevent the $out`/bin` to be a symlink
```

GTrunSec:
It's working!
image

- also double checked  the `nix.conf`

```
cat /etc/nix.conf
sandbox = false
accept-flake-config = true
experimental-features = nix-command flakes
```

PR #336, opened 2023-08-09
fix: prepare for potential future changes in drv output selection fragment

blaggacao:
# Context

On newer version of Nix (e.g. 2.16), this warning hints the user on potential upcoming changes:

```console
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '[ ... ].drv^*'
```

# Solution

This PR implements the suggestion of the warning message verbatim.


gytis-ivaskevicius (approved):
Havent tested it but looks right

blaggacao:
> assuming we consider nix v2.13 incompatible

Oh thanks for bringing this up. I hadn't considered this.

We could & probably should, meaning that while Paisano [requires 2.10.3](https://github.com/paisano-nix/core/blob/218a7f50451cbc5c981fc8dfe6f962044041f9aa/grow/default.nix#L211), Standard exceeds that and requires 2.13.

Will prep a fix.

blaggacao:
Tested and confirmed to implement the desired effect.

PR #344, opened 2023-08-24
FlakeHub: support uploading existing tags

grahamc:
If you're interested, you can publish existing tags using this updated workflow. Once this merges, you can go to the workflow on GitHub at https://github.com/divnix/std/actions/workflows/flakehub-publish-tagged.yml, click "Run workflow", and then type in -- for example -- `v0.24.0-1` or `v0.23.2`.

I hope that helps!

(also note that it'll automatically upload as divnix/std if it is published from divnix/std, no need to specify the `name` key if you don't want to 👍.)

### DIFF
--- a/.github/workflows/flakehub-publish-tagged.yml
+++ b/.github/workflows/flakehub-publish-tagged.yml
@@ -1,17 +1,26 @@
+name: "Publish tags to FlakeHub"
 on:
   push:
     tags:
       - "v*.*.*"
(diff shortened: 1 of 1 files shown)


blaggacao:
Oh nice, thank you. I did indeed wonder. For context: I just went through the click and run game coming from the FlakeHub lander because it was just so fast & convenient without much further thought.

PR #348, opened 2023-09-11
fix: broken `lib` follows

LiGoldragon:
I was unable to override the `nixlib` passed to inputs which used haumea's input.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -6,6 +6,7 @@
   description = "The Nix Flakes framework for perfectionists with deadlines";
   # override downstream with inputs.std.inputs.nixpkgs.follows = ...
   inputs.nixpkgs.url = "github:nixos/nixpkgs/nixpkgs-unstable";
+  inputs.lib.url = "github:nix-community/nixpkgs.lib";
   inputs = {
     paisano.url = "github:paisano-nix/core";
     paisano.inputs.nixpkgs.follows = "nixpkgs";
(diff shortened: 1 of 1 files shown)


whs-dot-hk:
You can override in `flake.nix` with `inputs.dmerge.inputs.nixlib.follows = "sth";`

blaggacao:
Ah just saw this. Yes, if I'd prefer if the direct second-level input override is workable, because this adds another (separate) input entry for `std` in downstream `flake.lock` files and `std` already suffers from excessive inputs, so if anything, I'd like to reduce them as much as we can (and keep them minimal).

whs-dot-hk:
O, sorry I actually mean `inputs.std.inputs.dmerge.inputs.nixlib.follows = "sth";`

LiGoldragon:
> O, sorry I actually mean `inputs.std.inputs.dmerge.inputs.nixlib.follows = "sth";`

That doesn't work:
```
 error: input 'std/dmerge/nixlib' follows a non-existent input 'std/haumea/nixpkgs'
 ```
 Even if `std.inputs.haumea.inputs.nixpkgs.follows` is set.

LiGoldragon:
Aditionally, the whole of avoiding to call `lib` by its long-established name is absurd. Nixpkgs is not lib.

blaggacao:
> Nixpkgs is not lib.

I agree. There's semantic value to this PR, regardless if it could be worked around. Maybe that should prime over reduction of inputs.

blaggacao:
adapted in https://github.com/divnix/std/pull/349

LiGoldragon:
> adapted in https://github.com/divnix/std/pull/349

You're awesome! 🙏

Issue #235, opened 2023-01-09
flake-module: defining `std.grow.nixpkgsConfig` should be optional


montchr:
As a potential user of the flake-parts module, I was surprised to find that setting the `std.grow.nixpkgsConfig` option is required, since I don't necessarily need to configure anything like `allowUnfree`.

blaggacao:
oops! It shouldn't be required.

blaggacao:
Thanks!

PR #354, opened 2023-09-16
Set default for nixpkgsConfig in flake-parts module

bnavetta:
For divnix/std#235, `std.grow.nixpkgsConfig` was made optional. However (at least in my config), `nixpkgsConfig` is not defined when it's checked for, even if it's set. Instead, I think setting a default of `{ }` should work - as far as I can tell, that's ultimately what it defaults to in nixpkgs.


blaggacao (approved):
Yep, that makes actually more sense!

PR #356, opened 2023-09-28
Fix kubectl arguments order

whs-dot-hk:
# Context
While running apply and diff through `std` cli,
`kubectl` will produce an unexpected args error.
# Solution
Put the filename arg after recursive,
so that it can be followed by a manifest path


blaggacao (approved):

blaggacao:
Thank's a lot! My use case / prototype had been focused on `kustomize` so I didn't catch that error. :handshake: Thanks for your collab!

Issue #357, opened 2023-10-08
The upgrade command from std TUI fails on the example flake for the `hello` package


blakat360:

whs-dot-hk:
I see, it is expected? You have to use `cargo new` to create your rust project

kolloch:
For me, pretty much all commands on packages except `build` fail due to quoting problems detected by shellcheck.

Attempted fix: https://github.com/divnix/std/pull/358

```
❯ std //hello/packages/hello:bundle
this derivation will be built:
nix-output-monitor error: DerivationReadError /nix/store/hqxv22vbvwy1nwwf6vp40i32sn37mpmk-source.drv: hGetContents: invalid argument (cannot decode byte sequence starting from 195)
  /nix/store/bdgmv51rwln9a9fcn0lz4flfwcnr8320-bundle.drv
bundle> building '/nix/store/bdgmv51rwln9a9fcn0lz4flfwcnr8320-bundle.drv'
bundle> 
bundle> In /nix/store/srzzp8mrqrb7f42klkxyj3v20jqap4id-bundle line 15:
bundle> nix bundle --bundler github:Ninlives/relocatable.nix --refresh $PRJ_ROOT#"x86_64-linux"."hello"."packages"."hello"
bundle>                                                                ^-------^ SC2086 (info): Double quote to prevent globbing and word splitting.
bundle>                                                                                        ^-- SC2140 (warning): Word is of the form "A"B"C" (B indicated). Did you mean "ABC" or "A\"B\"C"?
bundle>                                                                                                ^-- SC2140 (warning): Word is of the form "A"B"C" (B indicated). Did you mean "ABC" or "A\"B\"C"?
bundle>                                                                                                           ^-- SC2140 (warning): Word is of the form "A"B"C" (B indicated). Did you mean "ABC" or "A\"B\"C"?
bundle> 
bundle> Did you mean: 
bundle> nix bundle --bundler github:Ninlives/relocatable.nix --refresh "$PRJ_ROOT"#"x86_64-linux"."hello"."packages"."hello"
bundle> 
bundle> For more information:
bundle>   https://www.shellcheck.net/wiki/SC2140 -- Word is of the form "A"B"C" (B in...
bundle>   https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
error: builder for '/nix/store/bdgmv51rwln9a9fcn0lz4flfwcnr8320-bundle.drv' failed with exit code 1;
       last 10 log lines:
[...]

PR #358, opened 2023-10-23
fix: quoting of PRJ_PATH/fragment for installables

kolloch:
https://github.com/divnix/std/issues/357

### DIFF
--- a/src/local/flake.lock
+++ b/src/local/flake.lock
@@ -113,7 +113,10 @@
     },
     "haumea": {
       "inputs": {
-        "nixpkgs": "nixpkgs_2"
+        "nixpkgs": [
+          "std",
+          "lib"
+        ]
       },
       "locked": {
         "lastModified": 1685133229,
@@ -250,21 +253,6 @@
         "type": "github"
       }
     },
-    "nixpkgs_2": {
-      "locked": {
-        "lastModified": 1681001314,
-        "narHash": "sha256-5sDnCLdrKZqxLPK4KA8+f4A3YKO/u6ElpMILvX0g72c=",
-        "owner": "nix-community",
-        "repo": "nixpkgs.lib",
-        "rev": "367c0e1086a4eb4502b24d872cea2c7acdd557f4",
-        "type": "github"
-      },
-      "original": {
-        "owner": "nix-community",
-        "repo": "nixpkgs.lib",
-        "type": "github"
-      }
-    },
     "nosys": {
       "locked": {
         "lastModified": 1668010795,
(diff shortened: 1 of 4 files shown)


kolloch:
I also enabled showing the actual command with `set -x` that might not be to your test.

The precommit hooks in the devshell also put the updated flake.locks into this commit. Not sure about that.

whs-dot-hk:
I see the string created here https://github.com/paisano-nix/core/blob/3e897a19418361ece34841105122ed4f9379ca96/grow/newExtractFor.nix#L10

kolloch:
@whs-dot-hk I am not quite sure what you are getting it.

Maybe somemore context on what I did with the fragment:

I am quoting the fragment for shell usage. That would actually mean that the string including the double quotes is passed to the nix command.  Right now they are interpreted by bash, so that the string passed to the command does not contain the quotes anymore. That would make a difference, if anything in the path would contain spaces or other special characters, otherwise not so much. Still, I think it is generally good practice to quote things properly so that you don't even have to think about corner cases.

whs-dot-hk:
I was trying to understand what is fragment, it seems that it means a flake fragment

I am not sure where to escape, but I think yours is a good place

Do other places in the code also need to change?

kolloch:
> Do other places in the code also need to change?

Hard to say.

```
❯ rg PRJ_ROOT
src/std/fwlib/_mkCommand.nix
27:          if test -z "$PRJ_ROOT"; then

src/std/fwlib/blockTypes/terra.nix
69:        dir="$PRJ_ROOT/${repoFolder}/.tf"
70:        mkdir -p "$PRJ_ROOT/${repoFolder}/.tf"
71:        cat << MESSAGE > "$PRJ_ROOT/${repoFolder}/.tf/readme.md"

src/std/fwlib/blockTypes/installables.nix
37:        nix profile install $PRJ_ROOT#${fragment}
41:        nix profile upgrade $PRJ_ROOT#${fragment}
45:        nix profile remove $PRJ_ROOT#${fragment}
50:        nix bundle --bundler github:Ninlives/relocatable.nix --refresh $PRJ_ROOT#${fragment}
54:        nix bundle --bundler github:NixOS/bundlers#toDockerImage --refresh $PRJ_ROOT#${fragment}
58:        nix bundle --bundler github:ralismark/nix-appimage --refresh $PRJ_ROOT#${fragment}
```

`PRJ_ROOT` itself looks good to me.

Otherwise one would need to go through all instances in which nix variables are injected into shell scripts (that are not just store paths) and escape them.

blaggacao (approved):
That looks pretty sensible, indeed!

Let's wait for the test run.

PR #362, opened 2023-11-28
fix: Improve readability of terraform errors

dantefromhell:
When terraform errors output the affected line-of-code the output is highly unreadable as `config.tf.json' is a one-line JSON file.

Since there seems to be no nix expression for pretty printing JSON, this PR uses `jq` on the command line instead.

The result is that terraform syntax errors output the single affected line-of-code.


blaggacao:
Cool!! Thanks!

PR #369, opened 2024-02-17
feat: add nixostests block type

blaggacao:
# Context

- You may have a SDLC which runs nixos tests
- Especially if your deployment target is nixos

# Proposed solution

- Add nixostest block type
- Add iptable rules so that you can interact with the vm on your browser naturally


blaggacao:
mac failure unrelated snapshots pass.

PR #373, opened 2024-02-19
fix `go mod tidy` error with pinned PR version

nazarewk:
probably `nixpkgs` from `flake.lock` are just too old to work (almost exactly a year between nixpkgs versions)

I did 2 things:
1. `nix flake lock --update-input nixpkgs`
2. pinned to my own TUI from https://github.com/paisano-nix/tui/pull/8

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -115,11 +115,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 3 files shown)


nazarewk:
closing because it was merged by other means

PR #384, opened 2024-07-26
fix!: oci perms

blaggacao:
existing users of `config.copyToRoot` quite likely need to hove the contents of this key to `setup = [];` &mdash; `config.copyToRoot` may prime the nix2container permission cache with a parent folder (e.g. `/bin`) and conflicting permissions


blaggacao:
see: https://github.com/nlewo/nix2container/issues/144

Issue #386, opened 2024-07-29
Missing documentation for `blockTypes`


Lehmanator:
The following `blockTypes` don't have their definitions included in the documentation:

- `files`
- `kubectl`
- `microvms`
- `namaka`
- `nixostests`
- `nomad`
- `nvfetcher`

Documentation for how to define targets and their attribute layouts is missing for all `blockTypes`. Minimally, it would be nice to have at least one working example of a target for each `blockType`.

PR #388, opened 2024-07-30
docs: Add sections for missing `blockTypes`

Lehmanator:
Add sections to "Builtin Block Types" for each `blockType` without one.

Fixes: #386 (partially)

- Matches existing sections by inserting the contents of the file with the `blockType` definition.
- Fixes typo in comment in `blockTypes.nvfetcher` definition file.

If desired, I'd also like to:

- [ ] Alphabetize the `blockType` sections (or order according to another scheme) 
- [ ] Remove the comment strings from the `blockType` definition files and add them to the documentation.
- [ ] Create example definition files for each `blockType` and add them to the docs.

I could do this in another PR if out of scope for this one.

### DIFF
--- a/docs/reference/blocktypes.md
+++ b/docs/reference/blocktypes.md
@@ -82,3 +82,68 @@ _Note: while the implementation is the same as `functions`, the semantics are di
 ```nix
 {{#include ../../src/std/fwlib/blockTypes/anything.nix}}
 ```
+
+## Kubectl
+
+Block type for rendering deployment manifests for the [Kubernetes] Cluster scheduler.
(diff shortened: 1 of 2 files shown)


blaggacao:
> Alphabetize the blockType sections (or order according to another scheme)
:+1: 

> Remove the comment strings from the blockType definition files and add them to the documentation.
:-1: There are indeed folks for whom the documentation is secondary, think of workflows using lsp "jump to definition", even if not currently possible with nix, approximatable with `rg`/`grep`

> Create example definition files for each blockType and add them to the docs.

:+1: :100:

Issue #393, opened 2024-10-16
`src/data` using non-overridable `flake.lock` holding nixpgks from the distant past


nazarewk:
So I tried to reason about those warnings:

```
trace: warning: getExe: Package "conform-0.1.0-alpha.27" does not have the meta.mainProgram attribute. We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold. If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away. Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar".
trace: warning: getExe: Package "treefmt-0.5.0" does not have the meta.mainProgram attribute. We'll assume that the main program has the same name for now, but this behavior is deprecated, because it leads to surprising errors when the assumption does not hold. If the package has a main program, please set `meta.mainProgram` in its definition to make this warning go away. Otherwise, if the package does not have a main program, or if you don't control its definition, use getExe' to specify the name to the program, such as lib.getExe' foo "bar".
```

I did not have ANY reference to such `nixpkgs` definition in my flake code, then looking around `lib.getExe` usages only 2 candidates are inside https://github.com/divnix/std/blob/f8f6f70cdc9234d36d3d445d99a60b9267644df8/src/data/configs/lefthook.nix#L12-L12

then I have noticed `flake.nix` and `flake.lock` lying around in `src/data/` referencing [`nixpkgs` commit holding the faulty `treefmt-0.5.0` ](https://github.com/NixOS/nixpkgs/blob/f2ea252d23ebc9a5336bf6a61e0644921f64e67c/pkgs/development/tools/treefmt/default.nix)

I have no idea how and why this happens, but seems like `src/data` is loading and using packages from nixpkgs dated at 2023-09 with no way to modify this.

nazarewk:
sounds like `paisano` is trying to load cell blocks as flakes [in here](https://github.com/paisano-nix/core/blob/adcf742bc9463c08764ca9e6955bd5e7dcf3a3fe/grow/default.nix#L91-L100), I don't think this is a good idea for the default/undocumented behavior.

nazarewk:
seems like I have triggered the issue myself with https://github.com/divnix/std/pull/374 unaware of how it works underneath

blaggacao:
[...]

PR #394, opened 2024-10-17
fix: src/data:  #393 by extracting mdbook to own cell

nazarewk:
also switched `std` to nixpkgs on `nixos-unstable` and applied a few repo tweaks


nazarewk:
not sure how to properly generate those failing lockfiles

blaggacao (changes requested):
This looks great. However could we stay on a release branch for the main dependency?

We'll need to figure out why the manual re-locking is now failing in CI.

src/lib/cfg/lefthook.nix
```diff
@@ -5,7 +5,7 @@ let
   mkScript = stage:
     nixpkgs.writeScript "lefthook-${stage}" ''
       #!${nixpkgs.runtimeShell}
-      [ "$LEFTHOOK" == "0" ] || ${lib.getExe nixpkgs.lefthook} run "${stage}" "$@"
+      [ "$LEFTHOOK" == "0" ] || ${lib.getExe nixpkgs.lefthook} run --no-auto-install "${stage}" "$@"
```
nazarewk, line 8: actually we could switch to 24.05 since it contains this flag implemented

nazarewk:
> We'll need to figure out why the manual re-locking is now failing in CI.

It's actually pretty obvious and I don't know why it worked in the first place:
1. you update `/src/local/flake.lock` to match the new `../..`
2. `../..` flake just changed so you need to update `/src/local/flake.lock`
3. go back to 1.

blaggacao:
I made sure tests run on `main`: https://github.com/divnix/std/actions/runs/11517292880 &mdash; can you rebase?

PR #396, opened 2025-02-17
fix: prettier toml plugin path

womfoo:
make `prettier-plugin-toml` work again


blaggacao:
Ah nice, thank you! :smile:

PR #400, opened 2025-03-31
fix: balance nvfetcher script quotes

licebmi:
As stated on the documentation.

> '' is escaped by prefixing it with one single quote (')


blaggacao:
Thanks!

PR #401, opened 2025-05-14
Disable errexit for kubectl diff

whs-dot-hk:
Greetings @blaggacao, find the github actions exit unexpectedly after running the kubectl diff command.
# Problem
When there is difference `kubectl diff` will return exit status 1. If we have `set -o errexit` the script will exit after that command.
# Solution
Wrap `kubectl diff` command with following
```sh
set +o errexit
kubectl diff ...
set -o errexit
```


blaggacao:
Nice catch! Thank's a ton! :handshake:

PR #418, opened 2025-08-25
refactor: change subflake locking to an absolute git locking strategy for compatibility with nix 2.18+

### LINKED ISSUE #411: Update nix version in CI to min 2.18
Currently CI fails because the nixpkgs version requires nix version 2.18.

We should update CI accordingly.

### LINKED ISSUE #416: Move away from the self-referential locking in sub-flakes
After updating to 2.18 (or above), CI fails with

> `error: in pure evaluation mode, 'fetchTree' will not fetch unlocked input 'path:../../?lastModified=1'`

**Context**

- In this repository, we early attempted to use sub-flakes to sparsely inject inputs only to the (sub-)flakes, where they where needed.
- This allowed subflakes to access newer versions of an input that was not yet suitable for the main flake (e.g. due to stability reasons).
- It also allowed the subflake to load inputs that should not have increased the lock file of the main flake, because lock file size is of concern for a library like Standard. On a side note, this is also way some flake inputs are "blanked" and why there is a user-facing alert mechanism to adivse on adding these inputs in paisano-nix should the user wish to use certain functionality

However, it appears that the use of sub-flakes has been made completely impractical as of Nix v2.18. It was always a bit of an exotic use and Nix didn't do the "right" thing (imho) to maintain the trust scope of the git repository and render relative links between flakes _within_ that repo boundary trusted by default (meaning to waive the need for a lock or hash).

This repo hacked around this by choosing a sequence of locking the main flake and then re-locking the subflake with the lock of the commited main repository and thereby feeding the required lock of the anyways trusted own git repository into the subflake.

We should investigate how this evolved and is done today, then implement a new and more future-proof mechanism and ultimately leave this "hack" behind.

blaggacao:
Fixes #416 - alternative to #417.

Fixes https://github.com/divnix/std/issues/411.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,11 +1,8 @@
 #!/bin/sh
 
-# first time
-if [[ ! -d $(nix store add-path --name source --dry-run .) ]]; then
-  nix store add-path --name source .
-  (cd ./src/local && nix flake lock --update-input std)
-  (cd ./src/tests && nix flake lock --update-input std)
-fi
+# This locks subflakes to the latest commit of std
+(cd ./src/local && nix flake lock --override-input std git+file://$(pwd)/../..?rev=$(git -C ../.. rev-parse HEAD))
+(cd ./src/tests && nix flake lock --override-input std git+file://$(pwd)/../..?rev=$(git -C ../.. rev-parse HEAD))
 
 # shellcheck disable=SC1090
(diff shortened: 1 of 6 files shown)


DrTeagle:
I think this merits a fuller discussion(I understand both sides) of if we should more fully switch to flake-parts. It seems that they already have a flakePartitions module similar to what this does, except it works with newer nix for reasons I do not know.

PR #419, opened 2025-09-15
Fix prettier

DrTeagle:
This is a heavy handed approach and is very ... naive, it just removes all mentions of prettier-toml.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -70,7 +70,7 @@
       inherit (fwlib') blockTypes actions dataWith flakeModule grow growOn findTargets;
     };
   in
-    assert inputs.nixpkgs.lib.assertMsg ((builtins.compareVersions builtins.nixVersion "2.13") >= 0) "The truth is: you'll need a newer nix version to use Standard (minimum: v2.13).";
+    assert inputs.nixpkgs.lib.assertMsg ((builtins.compareVersions builtins.nixVersion "2.18") >= 0) "The truth is: you'll need a newer nix version to use Standard (minimum: v2.18).";
       (import ./dogfood.nix (inputs
         // {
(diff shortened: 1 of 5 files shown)


DrTeagle:
oh shit this has one more commit than i thought it did, whoops. Imma have to drop the more recent commit.

oneingan:
the interface will change upstream: https://github.com/NixOS/nixpkgs/pull/442943

 ```nix
  pkgs.prettier.override {
    plugins = with pkgs.prettier.plugins; [
      prettier-plugin-php
      prettier-plugin-pug
      prettier-plugin-ruby
      prettier-plugin-xml
      # ...
    ];
  }
  ```

much more convenient

DrTeagle:
I think this might depend on the referential flakes being resolved, as we can't use that nixpkgs until we can up our nix for CI(?).

DrTeagle:
I think that the mentioned PR wont be merged soon enough.

oneingan:
we must move to taplo

DrTeagle:
fair, is that the only place that prettier is used here for? toml? or is taplo more general than I recall. Also I am probably not gonna do much more for this PR, and genuinely forgot about it, I tried using this + hive for my nixos config, and got stuck on stylix overlays or something? That is to say, I am no longer using this in any capacity. I do think this project is a good idea, and I like the approach of standardization; hell, the idea of making there be features that can only be turned on by overriding inputs for other flakes is a genuinely good idea I believe.

That said I am closing this PR, and unless someone reaches out to me I am not going to consider making a PR here, as I both 
1. Don't have the time
2. Don't have the experience

I am in a rut in life as well so this position will probably not change soon enough.

PR #423, opened 2026-02-26
terra: read wrapper settings from terranix _meta

oneingan:
This lets the `terra` blocktype consume optional wrapper/backing settings from terranix metadata, without changing the rendered Terraform JSON.

- New (optional) contract consumed from terranix: `_meta.std.*`
  - `_meta.std.package`: terraform/opentofu package (optionally already `withPlugins`)
  - `_meta.std.providers`: list of provider packages; when set we call `package.withPlugins`
  - `_meta.std.modules`: attrset of vendored module directories (symlinked into the staging dir)
  - `_meta.std.terraformBackendGit.*`: override the default `terraform-backend-git` wrapper; `enable = false` runs the CLI directly

Dependency
- Uses terranix `_meta` passthru support: https://github.com/terranix/terranix/pull/151
  - Without `_meta` support, `terra` falls back to current behaviour (terraform + terraform-backend-git with the blocktype repo).

Testing
- `nix flake check -L`

### DIFF
--- a/src/std/fwlib/blockTypes/terra.nix
+++ b/src/std/fwlib/blockTypes/terra.nix
@@ -18,6 +18,12 @@ Available actions:
   - state
   - refresh
   - destroy
+
+Optional terranix `_meta` passthru (requires terranix PR #151):
+  _meta.std = {
+    package, providers, modules,
+    terraformBackendGit = { enable, repo, ref, state; },
+  }
 */
 let
   inherit (root) mkCommand;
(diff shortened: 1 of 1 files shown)


oneingan:
Update: removed the immutable root module machinery for now and kept this PR focused/minimal.

- `terra` only consumes optional `_meta.std.{package,providers,modules,terraformBackendGit.*}`
- no hard asserts; defaults preserve current behaviour when `_meta` is absent

